PR #293, opened 2015-06-24
RHD-521 - ajax-loader.gif not found on DevOps Buzz page

Dantheman720:
../ only works for non-nested files.  Changed it to relative to base_url.


Dantheman720:
Opened in the new repo

PR #295, opened 2015-06-24
RHD-510 Remove dashes from name URLs

Dantheman720:


paulrobinson:
I think it's just URLS a user would see that needs changing. So I wouldn't change 'stylesheets/_enterprise-java.scss'.

paulrobinson:
I also need a list of re-directs that need applying, once this PR is merged. They need to be in this format:

      http://developers.redhat.com/products/eap/get-started -> http://developers.redhat.com/products/eap/getstarted

I saw a spreadsheet that you shared with me, but it seems to have more changes in than are covered in this issue. The syntax is also not in a format that is going to be useful for eng-ops to setup the redirects.

Dantheman720:
I updated the spreadsheet.  If I misunderstood, please let me know.

paulrobinson:
We need to re-think this. I'm closing the PR, but please keep the branch for when we take this back up.

PR #297, opened 2015-06-25
Fix information about showed entries below paginated table. Closes #R…

lukas-vlcek:
…HD-532.

### DIFF
--- a/javascripts/developer-materials.angular.js
+++ b/javascripts/developer-materials.angular.js
@@ -233,6 +233,18 @@ dcp.filter('formatName',function(){
   }
 });
 
+/**
+ * safeNumber is an "ng filter" that accept string value
+ * and convert to number (using radix of 10). If parsing
+ * fails (NaN) then 0 is returned.
+ */
+dcp.filter('safeNumber', function() {
+  return function(input) {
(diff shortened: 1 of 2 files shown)


paulrobinson:
There was a build for this: http://developers-pr.stage.redhat.com/pr/297/build/588/. Not sure why the reporting on here was lost.

paulrobinson:
@wesbos can you review? I agree with @lukas-vlcek's comment that it's probably better to move the calculations into the controller. Thoughts?

wesbos:
:+1:  good to go from a code pov

javascripts/developer-materials.angular.js
```diff
@@ -243,6 +255,12 @@ dcp.controller('developerMaterialsController', function($scope, materialService)
     size : 10
   };
 
+  // Add Math object to $scope so we can use it directly in Angular expressions
+  // like: {{ Math.min(data.materials.length, paginate.currentPage * pagination.size) }}
+  // This might not be clean technique from Angular perspective (more clear would be
+  // to do all required calculations in controller and not the view)
+  $scope.Math = Math;
```
wesbos: Goign to fix this - Math should not be passed into the scope

PR #299, opened 2015-07-01
RHD-274 Added Diagram for DevOps Solution Get Started Page

adelasofia:


paulrobinson:
This needs to go on the get-started page, when it's created. Closing as we have nowhere to put this ATM.

PR #310, opened 2015-07-09
RHD-514 Fixed collapse point of side navigation

adelasofia:


paulrobinson:
See discussion on https://issues.jboss.org/browse/RHD-514

PR #311, opened 2015-07-13
Remove wrap to fix height issues in IE RHD-557

wesbos:


paulrobinson:
Asking Harry to review...

PR #312, opened 2015-07-14
added BrowserSync proxy + refresher RHD-558

wesbos:
Not sure if we want this in the main repo - thoughts?


paulrobinson:
Lets discuss this in our call later today.

PR #321, opened 2015-07-21
added external icon RHD-415

wesbos:
Adds an external link icon to external links

Might not work perfectly in staging since we do the "urlfix", which changes links from pr.developer.redhat.com to localhost, but that won't be an issue on production

This also includes a fix to the `urlFix` function which was flooding the console with errors when there was no link available.

### DIFF
--- a/_partials/developer-materials-results.html.slim
+++ b/_partials/developer-materials-results.html.slim
@@ -37,8 +37,9 @@
           / Title
           td.title
             p
-              a(ng-href="{{item.fields.sys_url_view | urlFix}}")
+              a(ng-href="{{item.fields.sys_url_view | urlFix}}" )
                 | {{item.fields.sys_title}}
+                i.fa.fa-external-link(ng-hide="item | checkInternal")
             span.contributor(data-sys-contributor="{{item.fields.sys_author}}")
               a.name
                 | {{item.fields.sys_author | name}}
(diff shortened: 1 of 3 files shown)


paulrobinson:
My concern with this approach is that it only applies to the learn page. What about other pages?

wesbos:
This applies to anything using the developer materials angular app - aren't the rest of the pages handled by awestruct?

paulrobinson:
Yes, most are handled by Awestruct. I thought we had other JS that displayed links. Maybe we don't. We can address that problem when it arrises.

paulrobinson:
See: http://developers-pr.stage.redhat.com/pr/321/build/625/products/eap/learn/#!project=eap&randomize=true

Notice these two: https://www.dropbox.com/s/rfra9c7ml194in3/Screenshot%202015-07-23%2011.25.26.png?dl=0. One has an offsite link, the other doesn't. I appreciate that the staging URL is causing an unexpected offsite link. However, I'd expect it to be consistently wrong. Maybe it's caused by the URL text being long?

paulrobinson:
Also, there seems to be a lack of padding between the link text and the offsite icon: https://www.dropbox.com/s/pr1rykp7t7a3z5e/Screenshot%202015-07-23%2011.26.47.png?dl=0. Would it look better with some spacing?

wesbos:
Will chat with you about this one tomorrow

paulrobinson:
@wesbos can you fix the padding, then we can merge.

PR #344, opened 2015-08-04
RHD-291 - Index 'OpenShift' videos for Learn page

Dantheman720:


paulrobinson:
This looks good, but make sure you only put OpenShift (Enterprise) videos in that Playlist. The OpenShift Online Product will be added later, at which point we can add the videos.

PR #352, opened 2015-08-06
Improved the clarity of the Aweplug section

paulrobinson:
@ollyjshaw made a good point that the commented out Aweplug lines were confusing. I've moved them into a comment and added a description to make this clearer.


paulrobinson:
@ollyjshaw can you review please.

ollyjshaw:
:+1: Nice and clear

PR #355, opened 2015-08-06
RHD-601 DevNation Videos: Fixed styles and made table mobile-friendly

adelasofia:


paulrobinson:
A great improvement, thanks!

PR #358, opened 2015-08-07
RHD-400 Often a download is triggered twice

paulrobinson:


paulrobinson:
This seems to work. Steps to reproduce:

1. Visit http://it-developers-pr.stage.redhat.com/pr/358/build/680/products/fuse/download/
2. Click 'login' in top right. This ensures you are logged into the staging build, for when you hack the redirect later.
2. Click on 'Full Zip (614mb)' and wait for the redirects and download to start (notice you get redirected to production).
3. Wait for 2 downloads to trigger, then cancel them to prevent confusion.
4. Edit the URL to have the 'http://it-developers-pr.stage.redhat.com/pr/358/build/680' prefix. This will cause the new get-started page to be used.
5. Observe that only one download is triggered.

Given the hacking around, it's hard to say for sure if this will work in production. I suggest the system is carefully reviewed after it goes live.

wesbos:
:+1: I can confirm this fix works for me

PR #368, opened 2015-08-21
RHD-470 Made products item in header navigation active on sub pages

adelasofia:


lkrzyzanek:
http://it-developers-pr.stage.redhat.com/pr/368/build/707/products/atomic/overview/ looks good.

PR #410, opened 2015-09-08
Updated the tests to reflect the addition of RedHat Mobile Platform

ollyjshaw:


paulrobinson:
It wasn't A-MQ that was added, it was Red Hat Mobile Platform. I don't think that matters though? Maybe fix the commit message though as it's confusing.

ollyjshaw:
will do

PR #411, opened 2015-09-08
RHD-632 Made Get-Started step boxes same height

adelasofia:


wesbos:
Good for now - would be ideal to use Flexbox next time so we can limit our dependency on JavaScript for  styles. I've got a set of videos if you are interested in learning >> http://flexbox.io

wesbos:
:+1: Good to merge

PR #412, opened 2015-09-08
Removing bundler tag

ollyjshaw:
We are having problems with the Gemfile.lock on the build servers. Currently they are running bundler version 1.10.4 or earlier. This causes a miss match with people running 1.10.5 or later. https://github.com/bundler/bundler/issues/3697

The real fix will be to upgrade all the bundles on the build servers, but this will fix it for now. I'll probably go through this with you on Wednesday @paulrobinson.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -353,5 +353,3 @@ DEPENDENCIES
   uglifier (~> 2.0.1)
   wraith (~> 1.3.0)
 
-BUNDLED WITH
-   1.10.6


paulrobinson:
@LightGuard do you have any thoughts on the right way to fix this? See: http://jenkins.mw.lab.eng.bos.redhat.com/hudson/job/developers.redhat.com/852/console. After every production build we are now seeing a modified Gemfile.lock which prevents a 'pull with rebase' from happening prior to the cdn.yml commit at the end of the build.

ollyjshaw:
@paulrobinson @LightGuard 
Running a `bundle install` should NOT modify the Gemfile.lock. However it will if you cross the 1.10.4 1.10.5 boundary which I did in a previous commit.

LightGuard:
Bundle install will update the Gemfile.lock if you use a range version
selector (which we're doing on pretty much everything to stay update with
the latest x.y branch) and a new version of that gem has been released. We
can certainly lock everything down to a specific version and update the
deps when needed (the java way of doing things).

On Tue, Sep 8, 2015 at 9:11 AM, redhat-developer-ci <
notifications@github.com> wrote:

> Can one of the admins verify this patch?
>
> —
> <https://github.com/redhat-developer/developers.redhat.com/pull/412#issuecomment-138595880>
> .
>



-- 
Jason Porter
http://en.gravatar.com/lightguardjp

LightGuard:
Looks like the change here was that there was a new version of bundler which was used to create the gemfile.lock file. Really not that big of a deal here.

ollyjshaw:
@LightGuard A bundle update will update the Gemfile.lock using range versions. A bundle install will install the specific versions of the gems in the Gemfile.lock. We're talking about the bundle install here.

ollyjshaw:
@LightGuard but yes, the issue here is the update to the bundler version and some unintended consequences that caught a lot of people out (including me!). Agreed not a big deal.

PR #423, opened 2015-09-14
RHD-684 Made event cards be the same height

adelasofia:


wesbos:
:+1: Good to merge

paulrobinson:
See <a href="https://issues.jboss.org/browse/RHD-684">RHD-684</a> where we are discussing this.

PR #431, opened 2015-09-17
RHD-685 - Removed logic to duplicate event if it carries over into ne…

Dantheman720:
…xt month


paulrobinson:
This is not the right approach. See JIRA issue for details.

PR #438, opened 2015-09-23
Dev ops phase1.2

ollyjshaw:
Allowing a full, reliable docker staging pr build


_config/site.yml
```diff
@@ -85,7 +85,21 @@ profiles:
...
       :cdn_out_dir: *dev_cdn_out_dir
       :cdn_version: *dev_cdn_version
-  
+
+  # Pull Request configuration
+  staging_docker_pr:
+    <<: *profile
+    dcp_base_url: http://searchisko:8080/
+      #TODO We should be able to use HOSTNAME
+    dcp_base_protocol_relative_url: //10.3.13.63:<%= ENV['SEARCHISKO_HOST_PORT'] %>/
```
LightGuard: This won't work, we can't hard code IPs here because they may not always be the same. Isn't there a SEARCHISKO_HOST_IP or similar environment variable we can use here?

_docker/awestruct/Dockerfile
```diff
@@ -6,6 +6,15 @@ ENV LANG en_US.UTF-8
...
 ENV LC_ALL en_US.UTF-8
 
+RUN yum install -y  \
+  pngquant
+
+RUN wget http://prdownloads.sourceforge.net/optipng/optipng-0.7.5.tar.gz?download
+RUN mv optipng* optipng.tar.gz
+RUN tar -xvzf optipng.tar.gz
+WORKDIR /tmp/optipng-0.7.5
+RUN ./configure && make && make install
```
LightGuard: Any reason not to combine more of the optipng downloading and building into a single step?

_docker/awestruct/Dockerfile
```diff
@@ -20,7 +29,9 @@ WORKDIR /home/awestruct
 COPY Gemfile Gemfile
 COPY Gemfile.lock Gemfile.lock
 COPY gemrc /home/awestruct/.gemrc
-RUN bundle install
+
+#parallel install of gems
+RUN bundle install -j 10
```
LightGuard: Awesome! Somehow that had always alluded me, no clue you could do that.

_config/site.yml
ollyjshaw: Agreed. We need the non Docker IP of the build machine. This is a bit tricky to get hold of. I'll check Jenkins and the Environments again. Failing that I'll put an environment variable on the build machine.

_docker/awestruct/Dockerfile
ollyjshaw: Nope, good idea.

PR #443, opened 2015-09-29
RHD-718 Removed broken link in Software Collections, Developer Toolse…

adelasofia:
…t and RHEL docs


paulrobinson:
Notice this in the JIRA issue: 'Before this is closed we need to remove the skips / ignores in the blinkr config'

PR #447, opened 2015-10-01
Feature/drupal archetypes fix

LightGuard:


LightGuard:
Looks like we may need to double check on that _brms-quickstarts to see if it's the correct version we want to track.

paulrobinson:
Can you improve the commit message: 'No clue what happened here'

paulrobinson:
What are the steps to verify that the 'duplicate archetype' fix works?

LightGuard:
No, I cannot improve the commit message as I really don't know what happened, maybe there was an update to that branch of the sub module we were tracking, I don't really know. If you don't want that commit, which is fine, just cherry pick the one with the fix, they're not related at all.

LightGuard:
Well, you could run the docker script and watch for errors from drupal. That's the only way I found it.

_brms-quickstarts
```diff
@@ -1 +1 @@
-Subproject commit 66df6d3774cf1c1e978a23ef0640d3e65d9688e0
+Subproject commit 7fb9b53309204cd3d2040306a5bfdbd95706ac63
```
paulrobinson: Looks like you are reverting the QS from BxMS-6.1.0.GA to BRMS-6.0.2.GA. Notice: https://github.com/jboss-developer/jboss-brms-quickstarts/releases.

I suggest you drop this change.

LightGuard:
rebased

PR #460, opened 2015-10-07
RHD-674 Fixed date representation of multi-day events

adelasofia:


paulrobinson:
The weight of the font on the month is inconsistent. See: https://www.dropbox.com/s/jbnv49z5gn08s9o/Screenshot%202015-10-08%2015.06.04.png?dl=0

Also, I think we need to try and improve on the inconsistencies between the single-day event and the multi-day events. Can you discuss this with @insectengine?

adelasofia:
@paulrobinson Just added style fixes after talking to James

PR #393, opened 2015-09-02
[READY FOR REVIEW] Ongoing blinkr fixes

LightGuard:

PR #466, opened 2015-10-13
Specific update to blinkr due to failures with our local development …

ollyjshaw:
…setup

This is a targeted update of blinker as opposed to a complete 'bundle update' carried out here https://github.com/redhat-developer/developers.redhat.com/pull/393. 

This change works with
```bash
'bundle exec ./control.rb -b --run-the-stack'
```

After https://github.com/redhat-developer/developers.redhat.com/pull/393 it was failing with ERR connection refused.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -20,9 +20,9 @@ GIT
 
 GIT
   remote: git://github.com/awestruct/awestruct.git
-  revision: 17ea5fc65f9364523569d3fa1dfb1085c75951be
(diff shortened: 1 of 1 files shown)


LightGuard:
It would be better if this tracked the latest released version of awestruct instead of git. Latest release is 0.5.6

paulrobinson:
@ollyjshaw is this still relevant?

ollyjshaw:
No, it's all working now.

PR #476, opened 2015-10-15
Cucumber daily build

ollyjshaw:


paulrobinson:
@LightGuard can you review?

paulrobinson:
@RubyGemTSL Can you review the changes around the tests.

LightGuard:
Looks fine to me. Really wondering if we want to build phantomjs though or if we should ship the binary in the repo for the image.

RubyGemTSL:
+1 on this - looks good.

@LightGuard, that would be possible if running on Mac, or Windows, but on Linux for the latest version of phantom you have to build from source.

LightGuard:
@RubyGemTSL Not if you build it on the target architecture once and the just copy that binary to the repo and add it into the docker image that way :)

ollyjshaw:
@lightguard. That sounds like a good idea. Would be a big boost. Also, we can have a docker image repo too. That would make things easier.

LightGuard:
We have an internal repo for all of Red Hat, but I'm not sure of the
current state of that.

On Thu, Oct 15, 2015 at 11:27 AM, Oliver Shaw <notifications@github.com>
wrote:

> @lightguard <https://github.com/lightguard>. That sounds like a good
> idea. Would be a big boost. Also, we can have a docker image repo too. That
> would make things easier.
>
> —
> <https://github.com/redhat-developer/developers.redhat.com/pull/476#issuecomment-148465906>
> .
>



-- 
Jason Porter
http://en.gravatar.com/lightguardjp

ollyjshaw:
Ok cool, might be one for a separate improvement.

PR #477, opened 2015-10-15
VWO / ios9 content blocker - page still loads when content blockers are turned on RHD-850

wesbos:
The way VWO works is that it temporarily hides the page while it does any A/B testing changes. 

Problem is that the code to show the site again after the edits are made is blocked, and thus never loaded. 

VWO also has a 2 second fallback to just show the regular site, but this was never running because the code to set "finished" to true was in the blocked script. 

This update forces the original site to show after a timeout of 2 seconds.


paulrobinson:
Thanks Wes. How did you find out about this solution? Was it recommended by VWO?

PR #485, opened 2015-10-19
Added new 'BPM Suite' intro album

paulrobinson:


eschabell:
Looks great, please push to production and thanks!

PR #495, opened 2015-10-21
Using a pre built binary of phantom js

ollyjshaw:
Using a binary of phantomjs so the install is quicker. We are not using any 2.0 specific features so we're fine. All the tests run.


ollyjshaw:
@LightGuard @paulrobinson 
Can we get this in before tomorrow am (22/10/2015) to speed up things for everyone when we break out the docker setup.

LightGuard:
We're going to check if we need the standard c++ libraries and the development headers for this to run. If we don't Olly will update the pull request. If we do need them we're good to go with merging this in.

LightGuard:
:+1:

PR #504, opened 2015-10-22
RHD-934 - Added ability to add tag to base.html.slim for protecting j…

Dantheman720:
@paulrobinson I added in the livestreams that we will want to protect.  I hope I understood this correctly, but it should add data-tag="protected" to anything identified in protected.yml.


paulrobinson:
Thanks Dan,

I've carried on this work under https://github.com/redhat-developer/developers.redhat.com/pull/505

PR #506, opened 2015-10-23
RHD-934 - Removed the for loop to reduce the complexity.

Dantheman720:
I realized now that I cast the array to a string before.  I think this should now remove some of the complexity since the for loop doesn't exit upon finding the item in the array.  I believe include? will exit if the item is in the array


paulrobinson:
retest this please

Dantheman720:
Can one of the admins verify this patch?

PR #521, opened 2015-10-27
Unit tests, guard and Rake

ollyjshaw:


paulrobinson:
This works on my laptop.

paulrobinson:
This works on my laptop.

PR #529, opened 2015-10-30
wait_for_ajax to complete was not working, causing random tests to fail

RubyGemTSL:


Dantheman720:
Can one of the admins verify this patch?

PR #549, opened 2015-11-05
Allow smoke tests to be run against docker stack

ollyjshaw:


Gemfile.lock
```diff
@@ -397,3 +400,6 @@ DEPENDENCIES
   therubyracer
   uglifier (~> 2.0.1)
   wraith (~> 1.3.0)
+
+BUNDLED WITH
+   1.10.6
```
paulrobinson: I think you need to remove this change, right?

paulrobinson:
@LightGuard can you review this please?

.awestruct_ignore
```diff
@@ -13,3 +13,6 @@ add-your-event.html.slim
 stylesheets/fonts/SIL Open Font License.txt
 stylesheets/rh-icons
 CONTRIBUTING.md
+_docker
+_cucumber
+cucumber_failures
```
LightGuard: `cucumber_failures` needs to be added, but `_docker` and `_cucumber` don't need to be. They start with `_` which is auto-ignored by awestruct.

Gemfile.lock
LightGuard: Why would this need to be removed?

README.md
```diff
@@ -140,6 +140,10 @@ To run the acceptance tests (where HOST_TO_TEST is the host you are interested i
 
 `bundle exec ./control.rb ---acceptance_test_target[=HOST_TO_TEST]`
 
+To run the acceptance tests against the locally running docker stack:
+
+`bundle exec ./control.rb ---acceptance_test_docker`
```
LightGuard: I think you have an extra `-` in there, should only be two, not three, same with line 141.

Rakefile
```diff
@@ -106,7 +106,9 @@ end
 
 desc 'Build and preview the site locally in development mode'
 task :preview, [:profile] => :check do |task, args|
-  run_awestruct "-P #{args[:profile] || 'development'} -a -s --force -q --auto --livereload -b 0.0.0.0"
+  profile = args[:profile] || 'development'
+  awestruct_running_port = awestruct_port(profile)
+  run_awestruct "-P #{profile} -a -s --force -q --auto --livereload -b 0.0.0.0 -p #{awestruct_running_port}"
```
LightGuard: We should change `-b 0.0.0.0` to `-b localhost`

Rakefile
```diff
@@ -421,6 +423,15 @@ def which(cmd, opts = {})
   return $awestruct_cmd
 end
 
+def awestruct_port(profile)
+  if profile.to_sym==:docker
```
LightGuard: Please follow the whitespace around operations like the rest of the file

LightGuard: Also what happens if we use drupal? That should be added as well.

_cucumber/features/support/env.rb
```diff
@@ -18,7 +18,13 @@
 SCREENSHOT_DIRECTORY = '_cucumber/screenshots'
 
 Capybara.configure do |config|
-  host_to_test = ENV['HOST_TO_TEST'] ||= 'http://docker:32768'
+
+  if ENV['HOST_TO_TEST']== ''
```
LightGuard: `== ` => ` == ` please

Gemfile.lock
ollyjshaw: As long as the message is in full, this is fine.

Rakefile
ollyjshaw: This is not a change I am bringing in. Ok if we look at this in a separate change outside of this PR?

ollyjshaw:
retest this please

Rakefile
LightGuard: Sure, no problem

ollyjshaw: Thanks, I'll make a note to do it now.

ollyjshaw:
retest this please

PR #551, opened 2015-11-06
Adding a  `/register` endpoint to redirect to the register or login pages RHD-977

wesbos:


register.html.slim
```diff
@@ -0,0 +1,28 @@
+html
+head
+  javascript:
+    console.log("head started");
```
paulrobinson: Debug

wesbos: Sorry - forgot to hit save before my commit.

PR #564, opened 2015-11-12
WIP Dont merge RHD-971 Added octokit for git interactions

ollyjshaw:


RubyGemTSL:
@ollyjshaw I closed PR for RHD-1007 so dont worry, I'll reopen when this gets merged. (work for RHD-971 was moved to a separate ticket)

PR #578, opened 2015-11-17
fixing broken images on local builds

wesbos:


paulrobinson:
@LightGuard can you take a look at this. I'm not sure this is the right fix though.

I think the problem is that Wes and Adela are seeing missing images on local (non-docker) builds when live reload is enabled. Given the usefulness of live reload, I'm not sure we want to disable it. Probably better to fix the missing image problem.

LightGuard:
You're probably right, but it'll take an awestruct fix. I'm not sure what's
going on, so I'd have to dig. Something else to put in the priority queue.

On Wednesday, November 18, 2015, Paul Robinson <notifications@github.com>
wrote:

> @LightGuard <https://github.com/LightGuard> can you take a look at this.
> I'm not sure this is the right fix though.
>
> I think the problem is that Wes and Adela are seeing missing images on
> local (non-docker) builds when live reload is enabled. Given the usefulness
> of live reload, I'm not sure we want to disable it. Probably better to fix
> the missing image problem.
>
> —
> <https://github.com/redhat-developer/developers.redhat.com/pull/578#issuecomment-157650806>
> .
>


-- 
Sent from Gmail Mobile

PR #583, opened 2015-11-19
Default rake task should not be overriden

ollyjshaw:


paulrobinson:
What's the motivation for this? @LightGuard can you review please.

LightGuard:
Should we set a default rake task for this script?

PR #603, opened 2015-12-02
DONT MERGE Rhd 890 test fix

RubyGemTSL:


RubyGemTSL:
@adelasofia @paulrobinson @ollyjshaw  - I have emailed with information on what Adela needs to do in order to get this passing.

ollyjshaw:
We don't need this any more @adelasofia has the change. cc @RubyGemTSL @paulrobinson

PR #610, opened 2015-12-03
First attempt at using Status API for blinkr

LightGuard:
I didn't have a good way to push information back up into the wrap
method, so I did some c&p :(

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -364,23 +364,41 @@ end
 
 desc 'Run blinkr'
 task :blinkr, [:new, :pr_prefix, :build_prefix, :pull, :build, :verbose] do |task, args|
-  $staging_config ||= config 'staging'
-  base_path = "#{args[:pr_prefix]}/#{args[:pull]}"
-  base_url = "#{args[:new]}/#{base_path}/#{args[:build_prefix]}/#{args[:build]}/"
-  report_base_path = "#{base_path}/blinkr"
-  report_path = "#{report_base_path}/#{args[:build]}"
-  verbose_switch = args[:verbose] == 'verbose' ? '-v' : ''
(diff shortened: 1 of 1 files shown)


LightGuard:
Awesome! Looks like things are working well as you can see in the checks below :) Blinkr will now link to the job while it is building and when it is finished the `Details` link will go to the actual report.

paulrobinson:
Great!

@ollyjshaw can you do a code review please?

ollyjshaw:
It looks to be doing the job. It's prpbably better to use the existing pattern we have (my email was late sorry, but you should have it now). All the blinkr stuff looks great, I'd just reccomend wrapping the rake task as per here (https://github.com/redhat-developer/developers.redhat.com/blob/master/_cucumber/cucumber.rake#L39)

LightGuard:
@ollyjshaw I looked at that, and have since re-looked at it, but there isn't any way to change the target_url using the `wrap_with_progress` method after the task has run. There also isn't a way to get the result of a task execution once you invoke it, so there isn't any way to send things back up the invocation chain without using class variables in the Rake tasks, which imo isn't good practice for what we want to do here.

What I'd like to be done is to have the Detail link in the checks show the jenkins build url until the blinkr report is uploaded. At that point the check should be a success and the url should change to the new report url. I don't see the need to go through multiple clicks to get to the blinkr report, this would be a step backwards from what is currently done with the comments.

We'd need to really adjust how `wrap_with_progress` works (if it's even possible to do what I'd like to do here) to accomplish the same operations here.

ollyjshaw:
@LightGuard 

That sounds fair enough. It does need some work. I'll take these comments on board for next time.

PR #618, opened 2015-12-10
RHD-1250 Email confirmation from staging redirects to production

paulrobinson:
This is how we specify which URL to redirect to after email confirmation.


paulrobinson:
@ollyjshaw can you review?

paulrobinson:
This is failing as the PR job doesn't seem to cope with PRs on non-master branches (as we suspected).

ollyjshaw:
:+1:

PR #627, opened 2015-12-11
Using docker compose to run acceptance tests

ollyjshaw:


_cucumber/features/support/env.rb
```diff
@@ -20,7 +20,7 @@
 Capybara.configure do |config|
 
   if ENV['HOST_TO_TEST'] == ''
-    host_to_test = "http://docker:#{ENV['AWESTRUCT_HOST_PORT']}"
+    host_to_test = "http://developers.redhat.com"
```
paulrobinson: Did you mean to hard-code this URL in here?

paulrobinson:
@LightGuard can you review please? In particular, take a look at the changes to '_docker/drupal/Dockerfile'. Olly was having problems building the Drupal image.

Rakefile
```diff
@@ -408,10 +415,17 @@ def run_awestruct(args)
 
   if ENV['site_base_path']
     base_url = ENV['site_base_path']
-    base_url = "#{base_url}/#{ENV['site_path_suffix']}" if ENV['site_path_suffix']
+
+    if ENV['site_path_suffix'].to_s != ''
```
LightGuard: Could change this to a `chop.empty?`

Rakefile
```diff
@@ -408,10 +415,17 @@ def run_awestruct(args)
...
+      base_url = "#{base_url}/#{ENV['site_path_suffix']}"
+    end
   end
+
   args ||= "" # Make sure that args is initialized
-  args << " --url " + base_url if base_url
+
+  if base_url.to_s != ''
```
LightGuard: Same here

LightGuard:
It'll be fine, when I get my drupal stuff pushed I'll fix the drupal docker container.

Rakefile
ollyjshaw: The idea here is to protect against nils and empty strings. 

```ruby
ENV['NOT_THERE'].chop
```

Will give an ```NoMethodError: undefined method `chop' for nil:NilClass``` if that environment variable is nil.

To make sure it does't blow up we convert to string as, 
```ruby
nil.to_s
```
returns ```""```

I do think using ```empty?``` makes the intent more clear, so I'll make that change now thanks.

_cucumber/features/support/env.rb
ollyjshaw: Yes, if running through the control script then a host is normally set. If not we could baulk, or use the live site. What do you think?

ollyjshaw:
@paulrobinson let's try again

PR #629, opened 2015-12-14
RHD-1258 /register allways redirects to production

paulrobinson:


register.html.slim
```diff
@@ -8,7 +8,7 @@ head
       clientId: 'web'
     });
 
-    var redirectUri = 'http://developers.redhat.com/confirmation';
+    var redirectUri = app.ssoConfig.confirmation;
```
paulrobinson: @wesbos Any ideas on how I get the variables into this page? Currently I'm seeing 'Uncaught ReferenceError: app is not defined' when visiting this page.

I need to ensure the right KeyCloak server is used in staging.

I also need to paramaterise line 6 above (url: "https://developers.redhat.com/auth",)

wesbos: Are you loading all of our JS on that page?

wesbos: I see - let me fork this and see if we can just pull in those bits.

wesbos: pull this into yours https://github.com/redhat-developer/developers.redhat.com/pull/639

paulrobinson: I now get 'Uncaught ReferenceError: app is not defined'. Any ideas?

paulrobinson:
@wesbos can you take a look at why this is failing? See my comments above.

register.html.slim
wesbos: @paulrobinson You are getting this on staging?

wesbos:
@paulrobinson Can you cherry pick this into yours:

https://github.com/redhat-developer/developers.redhat.com/commit/34767ae777fdb9230ff10dad0bdbd6e114c20a81

paulrobinson:
Thanks. I wonder why I didn't do it that way the first time? Hopefully there's not something we are missing. I'll take a look.

ollyjshaw:
Needs a rebase

paulrobinson:
Replaced by https://github.com/redhat-developer/developers.redhat.com/pull/664

PR #636, opened 2015-12-15
RHD-278 Display Developer Blog posts on Product pages

paulrobinson:


paulrobinson:
retest this please

paulrobinson:
retest this please

javascripts/buzz.js
```diff
@@ -116,7 +116,11 @@ app.buzz = {
             }
             d.updatedDate = jQuery.timeago(new Date(d.sys_created));
             d.sys_description = d.sys_description[0].substr(0,197) + '...';
-            d.permanentLink = "//planet.jboss.org/post/" + d.sys_content_id;
+            if (d.sys_url_view.startsWith('http://developerblog.redhat.com')) {
```
paulrobinson: @wesbos any ideas why this fails? I'm trying to use: https://github.com/redhat-developer/developers.redhat.com/blob/master/javascripts/extensions.js#L18, but I get the error: 'Uncaught TypeError: d.sys_url_view.startsWith is not a function'

wesbos: `sys_url_view` is an array, not a string:



Change it to: `if(d.sys_url_view[0].startsWith(...)`

PR #640, opened 2015-12-16
Removed Google Experiments RHD-1277

wesbos:


paulrobinson:
This needs removing too: https://github.com/redhat-developer/developers.redhat.com/blob/master/_config/site.yml#L23

wesbos:
Done

PR #332, opened 2015-07-29
use REST API with download manager RHD-64

wesbos:
This is almost done - a few things that need to happen

1. The descriptions of the downloads need to be updated in the admin - @Dantheman720  can you help us out here?
2. Logic for when no downloads are available needs to be added - it's there but what to show needs to be added.

PR #657, opened 2015-12-23
[READY FOR REVIEW] Login/Register Rhd 970

RubyGemTSL:


RubyGemTSL:
retest this please?

RubyGemTSL:
retest this please?

RubyGemTSL:
retest this please?

ollyjshaw:
retest this please

RubyGemTSL:
retest this please?

_config/site.yml
```diff
@@ -56,6 +56,8 @@ profiles:
     dcp_base_protocol_relative_url: //dcpbeta-searchisko.rhcloud.com/
     dcp2_base_url: http://dcp2-searchisko.rhcloud.com/
     dcp2_base_protocol_relative_url: //dcp2-searchisko.rhcloud.com/
+    keycloak_account_url: https://it-developers.stage.redhat.com/auth/realms/rhd/account/
+    keycloak_auth_url: https://it-developers.stage.redhat.com/auth/
```
paulrobinson: You will also need to use the staging download manager, as they keycloak and DL-man need to talk to each other.

paulrobinson: And that is blocked on the merging: https://github.com/redhat-developer/developers.redhat.com/pull/332, which will hopefully happen this week.

paulrobinson:
@ollyjshaw can you review please.

paulrobinson:
I've had a quick look. Only issue I found is the one I mentioned above.

_cucumber/lib/pages/forgot_password_page.rb
```diff
@@ -0,0 +1,23 @@
...
+  element :confirmation, '#kc-feedback-text'
+
+
+  def initialize(driver)
+    super
+    loaded?('Forgot Password | Red Hat Developers')
+    wait_for_reset_password_page && wait_for_reset_password_form
+  end
+
+  def enter_username_or_email(email)
```
ollyjshaw: I think the method name is misleading here.

ollyjshaw:
:+1: from me. Just one minor comment.

_config/site.yml
paulrobinson: You can update the config to use the staging dl-man for all non-production configs.

RubyGemTSL:
@ollyjshaw can you have a quick look over my changes, please? If you're OK with them this can be merged.

ollyjshaw:
:+1:

paulrobinson:
The downloads are still going to production dl-man. This is why: https://github.com/redhat-developer/developers.redhat.com/blob/master/javascripts/downloads.js#L3. You can fix that by paramaterizing the url like this: https://github.com/redhat-developer/developers.redhat.com/blob/master/register.html.slim#L11

ollyjshaw:
retest this please

RubyGemTSL:
@paulrobinson the change you suggested (as discussed this morning) breaks the latest download link. Instead of 'Download latest' you see a 'Fetching latest downloads' link.

javascripts/downloads.js
```diff
@@ -1,6 +1,6 @@
 app.downloads = {};
 
-app.downloads.url = 'https://developers.redhat.com/download-manager/rest/available/';
+app.downloads.url = '#{site.base_url}/download-manager/rest/available/';
```
paulrobinson: Try:

      app.downloads.url = app.baseUrl + '/download-manager/rest/available/';

RubyGemTSL:
retest this please

RubyGemTSL:
@paulrobinson your suggestion should fix this, but its still failing - maybe this is related to https://issues.jboss.org/browse/RHD-1372

Is download manager down in staging? For example if I try https://it-developers.stage.redhat.com/download-manager/rest/available/ I get nothing returned.

paulrobinson:
I see the problem:

      The requested URL /pr/657/build/377/download-manager/rest/available/fuse was not found on this server.

Sorry, I wasn't thinking straight. You need to pass in the URL of the download-manager, not the base URL of the site. If there is no variable available in JS, then you can add it to namespace.js, as that file is templated by Awestruct. I can explain this to you in more detail if this doesn't make sense.

RubyGemTSL:
retest this please

RubyGemTSL:
@paulrobinson so I need to revert back to the initial URL?

paulrobinson:
More specifically. 

Add a 'download_manager_base_url' variable underneath every occurrence of the 'download_manager_file_base_url' variable in site.yml. Set it to '//developers.redhat.com/' or '//it-developers.stage.redhat.com' depending on wether staging or production DL-man is used.

Add the following to namespace.js:

      /*
       Download Manager Base URL
       */
      app.downloadManagerBaseUrl = "#{site.download_manager_base_url}";

And then change the URL to use the app.downloadManagerBaseUrl variable.

_config/site.yml
```diff
@@ -106,6 +112,8 @@ profiles:
     dcp_base_url: http://<%= ENV['STAGE_SEARCHISKO_HOST'] %>:<%= ENV['STAGE_SEARCHISKO_PORT'] %>/
     dcp2_base_url: http://dcp2-searchisko.rhcloud.com/
     dcp2_base_protocol_relative_url: //dcp2-searchisko.rhcloud.com/
+    keycloak_account_url: https://it-developers.stage.redhat.com/auth/realms/rhd/account/
+    keycloak_auth_url: https://it-developers.stage.redhat.com/auth/
     metrics: false
```
paulrobinson: you need to overide the dl-man URLs in here.

RubyGemTSL:
@paulrobinson - All checks have passed!

PR #659, opened 2016-01-05
Rhd 1254 b

paulrobinson:


_layouts/base.html.slim
```diff
@@ -13,6 +13,11 @@
       - if site.dtm_code
         / Adobe DTM staging
         script src="#{site.dtm_code}"
+
+    - if site.adobe_target
+      = javascripts("adobe-target", true) do
+        script src="#{site.base_url}/javascripts/adobe-target.js"
+
```
paulrobinson: @wesbos: This is being added at the end of the body. Do you know how I can add it to the top of the head?

paulrobinson:
Not sure this is the right approach. Will re-open when I know what's happening.

PR #660, opened 2016-01-05
RHD-1024 - Update hero promos on Events page

Dantheman720:


paulrobinson:
I'm pretty confident that we don't want the node.js banner in here as the event has passed. Can ypu drop it from the PR. we can re-add if I'm wrong.

events.html.slim
```diff
@@ -15,12 +15,8 @@ description: Discover the events and webinars featuring talks about JBoss produc
         .swipe-wrap
           .slide
             a(href="https://www.youtube.com/playlist?list=PLnLQldO10NUMqPvW5loz6aLGL3GrqJEGq")
```
paulrobinson: This should link to http://www.devnation.org/

PR #672, opened 2016-01-12
use current keycloak by inlining js RHD-1347 

wesbos:
Okay - the reason we were linking to a static version of keycloak is because our build process doesn't serve up individual JS files - just a bundle of everything. 

The register page only needs keycloak.js so a simple solution to that is to just inline the JS at build time.


paulrobinson:
This isn't working for me: http://it-developers-pr.stage.redhat.com/pr/672/build/371/register/

Notice the error in the JS console.

wesbos:
Hrm - not working like it did locally. Will have a look.

wesbos:
@paulrobinson Works now

PR #676, opened 2016-01-15
HOLD - RHD-1350 - Added in dcp2 logic for blogpost retrieval

Dantheman720:
Testing to make sure that this operates as intended in staging.  It worked alright locally, but the solution doesn't seem to be very efficient.  Will revisit this.

### DIFF
--- a/_layouts/base.html.slim
+++ b/_layouts/base.html.slim
@@ -255,6 +255,7 @@
       script src="#{site.base_url}/javascripts/carousel.js"
       script src="#{site.base_url}/javascripts/video.js"
       script src="#{site.base_url}/javascripts/downloads.js"
+      script src="#{site.base_url}/javascripts/vendor/moment.js"  
     - unless page.extra_javascripts.nil?
       - page.extra_javascripts.each do |id|
         = deferred_javascripts id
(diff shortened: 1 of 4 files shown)


paulrobinson:
This seems like a lot of code to do something, that on the surface seems quite simple. Might be worth checking with @wesbos, to see if there's a simpler way of doing it?

wesbos:
Looks like most of this was duped from the other DCP instance, yes?

Dantheman720:
Yes, much of the logic was duplicated from the developerMaterialsController.  So, my thinking was that with angular, controllers are injected when declared in the html (might not be correct).  Since we are doing an additional request with DCP2 followed by a merge and sort, I thought it would be best to create an additional controller so that the developer materials page wouldn't be bogged down and have unnecessary blogposts in the returned json.

My other thought was to create a service function that did the query, merge and sort logic and then add a conditional in the original developerMaterialsController to see whether or not we are on the homepage.  If we are on the homepage, then use the "getLatestMaterials" function instead of the usual "getMaterials".

I hope I haven't overcomplicated this too much.  Merging the two DCP instances isn't very difficult, my main concern is to minimize performance affects on other pages.

Dantheman720:
@wesbos @paulrobinson - I can make any changes you feel are necessary here.  Hopefully I didn't stray too far from the right solution here.  Do those ideas seem valid?

paulrobinson:
@wesbos and @Dantheman720, can you discuss this together? I'm not sure what the right approach is here.

paulrobinson:
Deferring this until DCP 2 is ready.

PR #678, opened 2016-01-15
adding fixes for failing production tests.

RubyGemTSL:
removed wip tag

removed unnecessary code.

### DIFF
--- a/_cucumber/features/support/hooks.rb
+++ b/_cucumber/features/support/hooks.rb
@@ -59,7 +59,7 @@ After do |scenario|
 end
 
 def user_logout
-  if Capybara.app_host == 'http://developers.redhat.com'
+  if Capybara.app_host == 'http://developers.redhat.com/'
     visit("https://developers.redhat.com/auth/realms/rhd/protocol/openid-connect/logout?redirect_uri=#{@redirect_url}%3Fredirect_fragment%3D!")
   else
     visit("https://it-developers.stage.redhat.com/auth/realms/rhd/protocol/openid-connect/logout?redirect_uri=#{@redirect_url}%3Fredirect_fragment%3D!")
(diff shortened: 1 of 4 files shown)


_cucumber/features/support/hooks.rb
```diff
@@ -59,6 +59,11 @@
 end
 
 def user_logout
-  visit("https://it-developers.stage.redhat.com/auth/realms/rhd/protocol/openid-connect/logout?redirect_uri=#{@redirect_url}%3Fredirect_fragment%3D!")
+  if Capybara.app_host.include?('stage') || Capybara.app_host.include?('docker')
```
ollyjshaw: I think what your are really doing here is testing if the environment is live? If that's the case, then test for the live URL as it is clearer and will allow us to use other forms of non-live

RubyGemTSL: No, this is a hack to logout. As we cant delete cookies using selenium.

RubyGemTSL: This was also part of a previous PR, I just added the production URL in this PR.

ollyjshaw: I'm suggesting that the code be changed to 

```
if Capybara.app_host.equals?('developers.redhat.com')
  visit("https://developers.redhat.com/auth/realms/rhd/protocol/openid-connect/logout....
else
  visit("https://it-developers.stage.redhat.com/auth/realms/rhd/protocol/openid-connect/logout....
```

logic is "for live, logout here. For everywhere else log out here"

RubyGemTSL:
retest this please

RubyGemTSL:
@paulrobinson just waiting for Blinkr then this can be merged? cc @ollyjshaw

ollyjshaw:
:+1:

PR #689, opened 2016-01-21
Added instructions for adding event header graphics

paulrobinson:


paulrobinson:
@insectengine can you review these instructions?

PR #693, opened 2016-01-22
Only wait for searchisko configure if we need to

ollyjshaw:


ollyjshaw:
@paul, if we bring this in I can complete the accompanying  jeknins job for https://issues.jboss.org/browse/DEVELOPER-2363

ollyjshaw:
@paulrobinson , if we bring this in I can complete the accompanying jeknins job for https://issues.jboss.org/browse/DEVELOPER-2363

PR #706, opened 2016-01-26
DEVELOPER-1803 - Split a smoke test and a full integration test suite

RubyGemTSL:


RubyGemTSL:
retest this please

RubyGemTSL:
@ollyjshaw @paulrobinson 

I've re-added the acceptance tests to the developers.redhat.com builder. I've modified the job to pass a CUCUMBER_TAGS=@smoke parameter so only scenarios tagged with @smoke will be ran.

http://jenkins.mw.lab.eng.bos.redhat.com/hudson/job/developers.redhat.com/configure-readonly/
and
http://jenkins.mw.lab.eng.bos.redhat.com/hudson/job/developers.redhat.com-frontend-acceptance-tests/configure-readonly/

_cucumber/cucumber.rake
```diff
@@ -27,7 +27,13 @@ task :features do |t, args|
     profile = 'parallel'
   end
 
-  exit_status = runner.run(profile)
+  if ENV['CUCUMBER_TAGS'].to_s.empty?
+    tags = nil
+  else
+    tags = ENV['CUCUMBER_TAGS']
```
ollyjshaw: I think it might be worth having some validation here. The only tags allowed are @smoke and @acceptance right?

ollyjshaw:
I think the validation is worth doing. Other than that it looks good

_cucumber/cucumber.rake
RubyGemTSL: You raise a good point there, as if there was a typo in the cucumber tag, for example @smke 0 tests would be executed with an exit code of 0, meaning they would pass. I've added a check in there was verifies the tag is within a known good tags array. New tags can be added or removed from there.

RubyGemTSL:
retest this please

RubyGemTSL:
@ollyjshaw if you're happy with the checks I added, @paulrobinson can merge this.

_cucumber/cucumber.rake
```diff
@@ -27,7 +27,17 @@ task :features do |t, args|
...
+    known_good_cucumber_tags = %w(@wip @smoke @cdk_beta @ignore)
+    unless known_good_cucumber_tags.include?(tags)
+      raise("#{tags} is not a known tag. If required it can be added to the known good cucumber tags in #{File.dirname(__FILE__)}/cucumber.rake!")
+    end
+  end
+
+  exit_status = runner.run(profile, tags)
   p "exit status was #{exit_status}"
```
ollyjshaw: Cool. As I see this now, only one tag can be passed in. Otherwise  known_good_cucumber_tags.include?(tags) would be strange things. As such I think we should rename CUCUMBER_TAGS to CUCUMBER_TAG

ollyjshaw: OK, so with multi tags the problem is this

irb(main):001:0> known_good_cucumber_tags = %w(@wip @smoke @cdk_beta @ignore)
=> ["@wip", "@smoke", "@cdk_beta", "@ignore"]
irb(main):002:0> known_good_cucumber_tags.include?("@wip @smoke")
=> false

RubyGemTSL: fixed this now @ollyjshaw - good spot... again!

RubyGemTSL:
@ollyjshaw can review latest change?

ollyjshaw:
:+1:

RubyGemTSL:
@paulrobinson Olly is happy with this now.

PR #716, opened 2016-01-29
DEVELOPER-2475 - Password reset Acceptance test is randomly failing causing unnecessary PR fails

RubyGemTSL:
fixing randomly failed test for forgot password.

adding wip

added more password reset users


RubyGemTSL:
@ollyjshaw doesn't need a review really but @paulrobinson may need you OK. I've added extra password reset users to avoid using the same 1 account all of the time which was causing issues on parallel runs. 

*It is not a long term fix, but reduces the issue occurring it in the short term until I can control creating users on the fly via the Keycloak API.

PR #733, opened 2016-02-02
DEVELOPER-1852 - BRMS and BPMSuite claim to only be available through…

Dantheman720:
… subscription

### DIFF
--- a/_layouts/product-get-started-old.html.slim
+++ b/_layouts/product-get-started-old.html.slim
@@ -46,9 +46,14 @@ br
           = page._extra_installation_instructions
     - else 
       li
-        asciidoc:
-          #{page.product.abbreviated_name} is only available through a full Red Hat subscription. Download #{page.product.abbreviated_name} by
-          https://access.redhat.com/jbossnetwork/restricted/listSoftware.html[logging in to the Red Hat Customer Portal]  and downloading it from there.
+        - if page.product.full_subscription_required
+          asciidoc:
+            #{page.product.abbreviated_name} is only available through a full Red Hat subscription. Download #{page.product.abbreviated_name} by 
+            https://access.redhat.com/jbossnetwork/restricted/listSoftware.html[logging in to the Red Hat Customer Portal] and downloading it from there.            
+
+        - else 
+          asciidoc:
+
     - if page.product.guides['Getting_Started_Guide']
       p
         | For more details, view the 
--- a/products/openshift/_common/product.yml
+++ b/products/openshift/_common/product.yml
@@ -4,6 +4,7 @@ product_logo: portal.png
 current_version: 2.2.0.GA
 customer_portal_download_url: https://rhn.redhat.com/rhn/software/channel/downloads/Download.do?cid=21355
 documentation_path: OpenShift_Enterprise
+full_subscription_required: true
 youtube_album: https://www.youtube.com/playlist?list=PLf3vm0UK6HKo9vXQ1Ywj5tfjCUsqyXBwq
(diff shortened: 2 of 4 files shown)


_layouts/product-get-started-old.html.slim
```diff
@@ -46,9 +46,14 @@ br
           = page._extra_installation_instructions
     - else 
       li
-        asciidoc:
-          #{page.product.abbreviated_name} is only available through a full Red Hat subscription. Download #{page.product.abbreviated_name} by
-          https://access.redhat.com/jbossnetwork/restricted/listSoftware.html[logging in to the Red Hat Customer Portal]  and downloading it from there.
+        - if page.product.id == 'brms' || page.product.id == 'bpmsuite'
```
paulrobinson: Isn't there a way of doing this without hard-coding these product codes into the template? Look at how other pages achieve this.

ollyjshaw:
retest this please

ollyjshaw:
retest this please

_layouts/product-get-started-old.html.slim
Dantheman720: Agreed.  I can add it product.yml.  That seems like the best place to put it.

ollyjshaw:
retest this please

_layouts/product-get-started-old.html.slim
paulrobinson: I'm still unsure why other pages are not susceptible to this problem. Any ideas?

Dantheman720: http://developers.redhat.com/products/datagrid/get-started/ has the same content.  It is generated from get-started pages that use the old layout product-get-started-old.

Dantheman720:
retest this please

ollyjshaw:
docker build

_layouts/product-get-started-old.html.slim
paulrobinson: OK, figured out the problem. See the line with 'if page.product.default_download_artifact'. This displays the offending text if that key is null. Now that we use a REST API to get 'default_download_artifact' it will always be null at page build time. I guess the new get-started pages are not susceptible to this. I think we need to figure out a better way, at build-time to decide wether to show this or not.

ollyjshaw:
retest this please

ollyjshaw:
retest this please

Dantheman720:
This is not an optimal solution, but it fixes the full subscription problem.

Could we switch this over from product-get-started-old.html.slim to product-get-started.html.slim.  8 products still use the old one, and this probably shouldn't be the case since we are using the REST api now.

Dantheman720:
retest this please

paulrobinson:
retest this please

products/bpmsuite/_common/product.yml
```diff
@@ -4,6 +4,7 @@ product_logo: bpm-suite.png
 has_installer: true
 current_version: 6.2.0.GA
 documentation_minor_version: 6.2
+no_full_subscription: true
```
paulrobinson: This is a bit confusing. I suggest you invert the flag and name it: 'full_subscription_required'. You'll need to update more .yml files though.

paulrobinson:
This one is still claiming a full subscription is needed: http://it-developers-pr.stage.redhat.com/pr/733/build/685/products/datagrid/get-started/

Dantheman720:
Will do.  Which products should have full subscription requirements?  I think that is the main thing I'm confused about.

paulrobinson:
It's basically those that don't offer a donwload link directly on their /downloads page.

PR #740, opened 2016-02-03
Cdk developer 2478 - remove hard coded community links

RubyGemTSL:


_cucumber/features/community.feature
```diff
@@ -9,10 +9,3 @@ Feature: Community Page Smoke Test
     Then I should see the Community page title
     And I should see at least "3" promoted links
     And I should see at least "70" community links
-
-  Scenario: Specific community pages
-    Given I am on the Community page
-    Then I should see some well known projects such as:
-      | Hibernate Tools |
-      | Infinispan      |
-      | Seam            |
```
paulrobinson: This test seems like it should pass. see: http://it-developers.stage.redhat.com/projects/. Why do we need to remove it?

PR #751, opened 2016-02-08
Update Searchisko config, prior to configuration

paulrobinson:


paulrobinson:
@ollyjshaw can you review?

The 'v2.1.2' branch of Searchisko *always* contains the configuration running in production DCP. Therefore we can ensure we have matching config, by updating the branch prior to configuration.

This is an improvement over the 'managing config versions' idea, as it doesn't require the version in the Dockerfile to be updated whenever the DCP config is updated.

PR #757, opened 2016-02-09
DEVELOPER-2520 - Remove Satellite and Atomic Host products from RHDev

Dantheman720:


paulrobinson:
You need to fix the download page tests too. See: https://github.com/redhat-developer/developers.redhat.com/blob/master/_cucumber/features/downloads.feature#L11

ollyjshaw:
retest this please

RubyGemTSL:
retest this please

Dantheman720:
@paulrobinson This created two broken links to atomic and satellite in devops/adoption.adoc.  I removed them from the second commit in case you would prefer to use a different change.

PR #758, opened 2016-02-09
DEVELOPER-1805 - Replace newline character with break for video descr…

Dantheman720:
…iptions


Dantheman720:
@paulrobinson - http://it-developers-pr.stage.redhat.com/pr/758/build/140/video/youtube/YoC0vDIWpC8/

This will render newline as an html break.  Is that the way we want to do it?  Currently, the newlines are represented in the full_description, it just doesn't translate to html very well.

PR #759, opened 2016-02-09
DEVELOPER-2523 and DEVELOPER-2524 (awestruct changes)

LightGuard:
This adds verbose logging to get (which I'm pretty sure is what we're doing in the builds) and also the performance logging, however I've not enabled the performance logging at all to do that you'd need to change the Rakefile anywhere you want the performance logging to be enabled and set the `run_awestruct` with the `--perf` flag. We'd also need to start saving the `.awestruct` directory in jenkins.


ollyjshaw:
:+1:

PR #769, opened 2016-02-11
DEVELOPER-2548 Switch staging URLs back to developers.stage.redhat.com

paulrobinson:
We'll also need to fix the URLs in the Jenkins config of the PR builders. Need to replace 'it-developers' with 'developers'.


paulrobinson:
retest this please

PR #771, opened 2016-02-11
Developer 2545 aweplug

ollyjshaw:


LightGuard:
There may be some things in aweplug you want to switch for this as well.

PR #790, opened 2016-02-15
Hard Launch changes, DEVELOPER-2530, and DEVELOPER-2526

adelasofia:


paulrobinson:
It doesn't look like the conflicts were resolved correctly. It's still claiming to have conflicts.

PR #808, opened 2016-02-18
[WIP] DEVELOPER-2524 Add performance execution logging to awestruct

paulrobinson:


paulrobinson:
@ollyjshaw can you confirm that i've added the '--perf' flag in the right place. I want it to trigger for this PR.

This is just a WiP so that I can see what the results look like. We will need a better way of enabling/disabling on demand.

PR #816, opened 2016-02-19
Switched CI code to expect cdn.store, not cdn.yml

paulrobinson:


paulrobinson:
@ollyjshaw, @LightGuard Can you review ASAP? Otherwise we need to backout Jason's YAML fix :-(

ollyjshaw:
:+1:

PR #825, opened 2016-02-22
Updated to new Searchisko config (v2.1.4)

paulrobinson:


ollyjshaw:
@paulrobinson Can we merge this? It's the only PR that currently builds. The others fail due to searchisko problems. I suspect this will fix everything else.

PR #856, opened 2016-02-29
make sure we expose the mysql port

ollyjshaw:


ollyjshaw:
@paulrobinson. This rectifies issues with telnetting to 3306 from searchisko

PR #871, opened 2016-03-02
Temporary switch to single-threaded builds

paulrobinson:
This is to work around a concurrency issue in the DCP.


paulrobinson:
retest this please

PR #893, opened 2016-03-08
DEVELOPER-2640 Implementation of RHEL Overview Page

adelasofia:


RubyGemTSL:
needs a rebase to get latest test updates with changes to login/register pages.

paulrobinson:
This was a pproved by @MikGue and the code looks good to me. Will merge when tests pass.

paulrobinson:
@RubyGemTSL can you confirm that this is failing for the same reason as the tests on the Octopus branch? If so, we can merge this as it doesn't make the situation any worse. We also need to get the Octopus branch tests passing ASAP. @RubyGemTSL is about to raise a PR for that.

RubyGemTSL:
@paulrobinson this is also fine to merge

PR #912, opened 2016-03-09
[HOLD] Groundwork for AB testing with Adobe Target

wesbos:
// To run - this code would go inside Adobe Target
// You can open your console and run it to try it out

```js
$(function() {
  app.abTest.swap('heros/hero2', '.homepage-main');
});
```

### DIFF
--- a/heros/hero1.html.slim
+++ b/heros/hero1.html.slim
@@ -1,0 +1,1 @@
+= partial('hero1.html.slim');
--- a/_layouts/base.html.slim
+++ b/_layouts/base.html.slim
@@ -321,6 +321,7 @@
       script src="#{site.base_url}/javascripts/video.js"
       script src="#{site.base_url}/javascripts/downloads.js"
       script src="#{site.base_url}/javascripts/latest.js"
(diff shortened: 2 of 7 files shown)


heros/hero1.html.slim
```diff
@@ -0,0 +1 @@
+= partial('hero1.html.slim');
```
paulrobinson: Why do you need this indirection? Can't you just put the contents of the partial inline?

paulrobinson: OK, I see now. I think you need the partial for the initial HTML build of the page, and then you use this file in the swap function. Is that right?

wesbos: Yep - we need the partials to be publicly available so we can pull them in
via the js if we are doing ab testing

paulrobinson:
Closing this PR, as we've confirmed that this is the right approach. Please include this code in your PR containing the real experiment. E.i https://issues.jboss.org/browse/DEVELOPER-2763 and https://issues.jboss.org/browse/DEVELOPER-2779

PR #913, opened 2016-03-09
DEVELOPER-2748 added new vJBug session

adelasofia:


RubyGemTSL:
retest this please

paulrobinson:
I noticed that the time seems 'too precise' on the two most recent past videos. Notice '02/09/2016 6:00:00 PM UTC' which would read better as '02/09/2016'

paulrobinson:
Also, the enlarged video doesn't display. See: https://www.dropbox.com/s/s83gzamjlbta7po/Screenshot%202016-03-10%2009.16.25.png?dl=0

It also doesn't fill the screen, on a large display. I think we've had that issue for a while though, so only fix here if its easy.

adelasofia:
@paulrobinson I've fixed the issue with the video not loading on the modal and the dates for the past sessions. I also created [this](https://issues.jboss.org/browse/DEVELOPER-2755) jira to fix the rest soon.

PR #916, opened 2016-03-10
Removed blocking Font awesome request DEVELOPER-2707

wesbos:
We are still loading font awesome via the CDN, but we just brought the CSS to load it into our own - this cuts down on 1 blocking request.


paulrobinson:
@LightGuard please can you review. I think you recently made a change in this area, to support the Drupal migration.

LightGuard:
You'll need to change 

- _drupal/themes/custom/rhd/rhd.libraries.yml 
- _drupal/themes/custom/rhd/rhd.info.yml

As well so we don't load font-awesome twice in drupal. The changes there will just be deletions.

wesbos:
@LightGuard @paulrobinson Okay - good to go now

PR #940, opened 2016-03-17
[Octopus] Merged Master into Octopus branch

adelasofia:


LightGuard:
This brings octopus up to date with the current changes in master.

paulrobinson:
This hack needs to be reverted for Octopus: https://github.com/redhat-developer/developers.redhat.com/pull/940/commits/2db6cc7e9aaf58762f1d34e70ed1960de38ce9d2

The hack was added to workaround the fact that production download manager knows about Octopus, but we don't want users linking to it ATM.

LightGuard:
@paulrobinson I would suggest that change be made in a different PR, this is simply bringing octopus up to date with master.

PR #942, opened 2016-03-17
DEVELOPER-2666 - update DTS and SCL download pages

Dantheman720:


Dantheman720:
This should be opened against Octopus since it incorporates the changes made to RHEL for the Octopus release.  See comments: https://issues.jboss.org/browse/DEVELOPER-2666

PR #970, opened 2016-03-30
DEVELOPER-2830 CDK2 TTHW Beta5 Updates.

rctrh:
* Beta5 updates for CDK2 TTHWs
* Changes are independent of the new CDK get-started page.  (Can be merged with or without)


paulrobinson:
this is ok to test

paulrobinson:
Looks like there is an error with file /products/cdk/get-started-cdk2-ruby.adoc. Maybe @adelasofia can advise?

rctrh:
@paulrobinson - This is the problem with Ruby code in a code block getting interpreted by something (awestruct?) that I mailed you about yesterday.

See commit 83d900e, I work around this by changing the Ruby code.

paulrobinson:
Yes, @adelasofia and I think so. Although we don't know for sure. @adelasofia can you verify?

LightGuard:
Are we okay to merge this or is work being done somewhere else? @adelasofia @rctrh

rctrh:
I think we are OK to merge this.  It can go out now, before the get-started page that @adelasofia is working on in #977

PR #971, opened 2016-03-30
[URGENT] DEVELOPER-2835 Added USB card to MS Build events page

adelasofia:


wesbos:
:+1: Code looks good to me

PR #974, opened 2016-03-31
DEVELOPER-2837: RHEL Get Started Alternate Post Install

rctrh:
Adds additional, temporary workaround instructions for RHEL post-install bug.

Note:needs to be reviewed in template.


paulrobinson:
this is ok to test

PR #976, opened 2016-04-01
DEVELOPER-2853 - Disallow backends in robots.txt

lkrzyzanek:


paulrobinson:
this is ok to test

lkrzyzanek:
Cann't verify because http://developers-pr.stage.redhat.com/pr/976/build/723/robots.txt has Disallow / for PR build. But since it doesn't break anything I think it's OK to merge.

PR #978, opened 2016-04-02
DEVELOPER-2859: RHEL TTHW Fixes: Java Typo, docker ruby missing line

rct:
* Delete extraneous / in Java code.
* Add back Ruby line that got deleted because it was being misinterpreted by awestruct (or some other Ruby component) that thought it was code instead of a literal per-formatted code block.


LightGuard:
this is okay to test

LightGuard:
ok to test

PR #991, opened 2016-04-06
Committed output of Gemfile.lock from running bundle install from com…

robpblake:
- Added Gemfile.lock from command line running of bundle install. Seems to add an extra new-line over what is checked in to master which should resolve  local modifications.


RubyGemTSL:
retest this please

RubyGemTSL:
@LightGuard can you merge this for us when you're online please?

PR #999, opened 2016-04-08
[DEVELOPER-2854] - Spike - Backup Drupal and Postgres

robpblake:
This PR is the result of the spike into backing up and restoring our Drupal and Postgres in our pilot environment.

The README.md provides commentary on how the backup/restore process works and how it could be integrated with Jenkins.

Changes to the Dockerfile for the Drupal image are to expose the Drupal filesystem as a volume to:

1. Ensure the filesystem survives accidental termination of the container instance
2. Avoid having to work with an overlay filesystem and instead write directly to the host filesystem for better performance
3. Allow the volume from the Drupal instance to be mounted onto the backup/restore container


robpblake:
@paulrobinson / @LightGuard I would appreciate your comments and feedback on this. 

This is the first version of the backup and restore process for the Drupal pilot environment. Please see the README.md for a full description of how it works and how it can be integrated with Jenkins.

LightGuard:
I'm going to have to take some time to go through this, hopefully later today.

LightGuard:
It seems sound :)

LightGuard:
I've switched over to mysql :) This may make things a little easier as you can do this from drupal console now:

```
drupal database:dump
drupal database:restore
```

Looks like we need to install mysqldump on the docker image to do this though.

robpblake:
This is no longer required. Closing PR.

PR #1018, opened 2016-04-14
Added Ian's new key to vault

paulrobinson:
@RubyGemTSL please download the secrets.yaml.gpg and check that you can build the site with it. If so you can review and merge this PR.


RubyGemTSL:
@paulrobinson @LightGuard It didn't work. I spoke to Rob - I did it wrong. I only generated a public key. I was informed I needed to generate a private one too. This can be closed.

PR #1019, opened 2016-04-14
DEVELOPER-2885 Articles in Drupal

LightGuard:
Layout files and small module additions within drupal

### DIFF
--- a/_docker/drupal/modules/custom/redhat_developers/config/install/filter.format.rhd_html.yml
+++ b/_docker/drupal/modules/custom/redhat_developers/config/install/filter.format.rhd_html.yml
@@ -3,7 +3,6 @@ status: true
 dependencies:
   module:
     - editor
-    - toc_filter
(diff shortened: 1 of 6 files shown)


LightGuard:
@Dantheman720 since you're the only other one who has it up and running on their box, would you mind taking a look at this and see if it matches the "Definition of Done", and following comments, in the ticket?

Dantheman720:
@LightGuard Will do.

Dantheman720:
The images currently are not being added to the body text. I think it would be easier for the user to add the image tags directly to the html so that users have the flexibility of adding the images wherever they want.  Other than that, I think it's pretty intuitive.

https://www.dropbox.com/s/e883atq5ke6n9c2/DrupalArticle.mp4?dl=0

LightGuard:
@adelasofia according to @Dantheman720 this is good to merge, mind handling the merge?

PR #1022, opened 2016-04-18
Remove "beta" from CDK2 info.

lincolnthree:
CDK 2.0 is releasing 2016-04-19 and will no longer be in Beta.


LightGuard:
ok to test

PR #1026, opened 2016-04-19
Link direct to .NET preview video

paulrobinson:
Now that the .NET webinar vide is live, I think we should link directly to it from the topic page, rather than linking to the webinar registration page.

@lincolnthree can you review please?


LightGuard:
Looks good to me

lincolnthree:
Approved.

PR #1027, opened 2016-04-19
DEVELOPER-2860 redirect on logout

LightGuard:
Changing the redirect URL on logout


LightGuard:
Hm... turns out this is harder to test than I thought because when you click a download link you don't go back to the URL you were on, but the staging site.... @paulrobinson any thoughts there?

javascripts/sso.js
```diff
@@ -15,7 +15,7 @@ app.sso = function () {
                 // once the promise comes back, listen for a click on logout
                 $('a.logout').on('click',function(e) {
                     e.preventDefault();
-                    keycloak.logout();
+                    keycloak.logout({"redirectUri":"https://developers.redhat.com"});
```
paulrobinson: I think this should use site.base_url to keep the user on the same environment.

LightGuard:
Tiffany says this is good to go.

PR #1029, opened 2016-04-19
DEVELOPER-2910 Updated footer

adelasofia:


adelasofia:
Approved by Tiffany N.

LightGuard:
@Dantheman720 We'll need to add this as well now to the drupal html.html.twig file

PR #1033, opened 2016-04-21
Update index.html.slim

insectengine:
Replace Devoxx France promo with SamSam update.


LightGuard:
ok to test

LightGuard:
@Dantheman720 We'll need to add this to the html.html.twig as well.

PR #1048, opened 2016-04-25
DEVELOPER-2951 switch to mysql for Drupal

LightGuard:
Switching back to MySQL for use with Drupal. There are a few changes in files, but they're all related.


paulrobinson:
@LightGuard Apparently it will be MariaDB 10.0.14 that we'l be provided with. Can you update from MySQL to a version of Maria DB close to 10.0.14? Sorry for messing around :-(

LightGuard:
@Dantheman720 could you verify this please?

Dantheman720:
+1 Looks good to me. I was able to add an article with no problems.

PR #1057, opened 2016-04-27
DEVELOPER-2914 Hid Devnation discount content before login

adelasofia:


adelasofia:
retest this please

adelasofia:
retest this please

adelasofia:
retest this please

RubyGemTSL:
retest this please

LightGuard:
Maybe I'm missing something, but um, is simply hiding it enough? Seems like we should be removing it.

adelasofia:
@LightGuard we are only hiding it until the user registers then it is shown

LightGuard:
Okay, fair enough

paulrobinson:
I think it would be better if the code was hidden by default. There's still a small amount of time when the page loads, that you can see it: https://www.dropbox.com/s/xk6aq4l6ceubilr/Screenshot%202016-04-28%2008.49.13.png?dl=0. Although, I'm not sure this is possible with slim?

I think we can merge this as an improvement over what is in production, once @MikGue approves.

promotions/devnation-discount-code.html.slim
```diff
@@ -20,9 +20,10 @@ drupal_type: rhd_microsite
         br
         | To register for DevNation 2016 and get the 30% discount, click
         a(href="https://rh2016.smarteventscloud.com/portal/createAccount.ww?")  here
-        |  and enter the passcode: “
-        strong DEVSAVE30
```
LightGuard: You could do:

    strong.hide DEVSAVE30

to hide it initially.

promotions/devnation-discount-code.html.slim
```diff
@@ -20,9 +20,10 @@ drupal_type: rhd_microsite
...
-        |”.
+        |  and enter the passcode: 
+        strong.devnation-code DEVSAVE30
```
LightGuard: You could use `.hide` on the `strong` element to hide it initially.

LightGuard:
Superseded by #1069

PR #1064, opened 2016-04-29
[URGENT] Switching from `present?` to core ruby methods

LightGuard:
I'm not sure why we switched to active_support's Object::present? method
but I'm switching back to core Ruby methods in an attempt to fix the
production builds.


paulrobinson:
Seems ok: http://developers-pr.stage.redhat.com/pr/1064/build/1067/video/youtube/i5V5kRFUyxI/

LightGuard:
Well, do we try pushing it then?

PR #1068, opened 2016-05-02
[DRUPAL] feature/DEVELOPER-2975 -- Idempotent pushes to Drupal

LightGuard:
Testing this may be a little difficult. You need a Drupal instance which already has content in it to see it work. Once you have that then you can simply run `bundle exec ./control.rb -u --run-the-stack`.

I think a possible way to test this on your own would be to modify the control.rb script so it doesn't kill or rebuild the containers.


paulrobinson:
Maybe one for @robpblake to review, given that he'll be working with this feature pretty soon.

robpblake:
I think the implementation needs to be unit tested. This is a key part of our delivery pipeline, so unit tests will make it easier for us to be confident that it works without having to spin up a complete stack and run end-to-end tests to make sure.

With regards to testing this myself, it is a little hard to follow the current build process so I'd like to double check my understanding:

The --run-the-stack (after some setup) I think finally comes down to the following command being run inside a launched container:

> rake git_setup clean gen(drupal)

This eventually calls out to:

> run_awestruct -P drupal -g --force -q -w

I believe this then generates the site from the content in the repository and pushes it into Drupal using the helper class above. 

My questions on this:

1. How and where is the Drupal endpoint into which content is pushed configured?
2. How is the above helper class configured into the chain of things for Awestruct "to do" when running the build of the site? Which method on the class is invoked by Awestruct to get the content pushed into Drupal?
3. If my understanding of the build process is correct, can we simply use the following to generate the content after starting a Drupal instance?

> bundle exec ./control.rb -g

LightGuard:
> I think the implementation needs to be unit tested. This is a key part of our delivery pipeline, so unit tests will make it easier for us to be confident that it works without having to spin up a complete stack and run end-to-end tests to make sure.

It'll take a little bit to actually get this tested, and we'd need some more dependencies, but that isn't much of an issue. I'm not sure if I'll be able to get it done in a timely manner this week though due to travel and the meeting in Raleigh

> With regards to testing this myself, it is a little hard to follow the current build process so I'd like to double check my understanding:

> The --run-the-stack (after some setup) I think finally comes down to the following command being run inside a launched container:

You need to use `-u` as well to tell the control script to start up drupal.

>    rake git_setup clean gen(drupal)
> This eventually calls out to:

>    run_awestruct -P drupal -g --force -q -w
> I believe this then generates the site from the content in the repository and pushes it into Drupal using the helper class above.

Correct.

> My questions on this:

> How and where is the Drupal endpoint into which content is pushed configured?

`_config/site.yml` There's a drupal profile in there which extends the docker profile.

> How is the above helper class configured into the chain of things for Awestruct "to do" when running the build of the site? 

It's instantiated in `_ext/pipeline.rb` as a transformer, it's near the bottom of the file.

> Which method on the class is invoked by Awestruct to get the content pushed into Drupal?

`transform(site, page, content)`

> If my understanding of the build process is correct, can we simply use the following to generate the content after starting a Drupal instance?
> bundle exec ./control.rb -g

If you use `-ug` yes, you should be able to, but right now it seems to kill any of the existing containers and restart them all, which isn't quite what we want.

LightGuard:
Tested, should be an integration test, but that's not going to happen without a full Drupal environment.

LightGuard:
@robpblake feel free to look over it

LightGuard:
retest this please

robpblake:
@LightGuard Looks good to me. 

@paulrobinson: Can you merge this if you're happy? (I don't have the privileges)

PR #1074, opened 2016-05-04
DEVELOPER-2950 - Drupal quickstart and demo pages are not working

Dantheman720:
Updates to incorporate aweplug changes for Drupal


Dantheman720:
@LightGuard - Can you review this when you get the chance?

LightGuard:
It looks okay from a code perspective, but the DCP died here again :(

LightGuard:
retest this please

LightGuard:
@Dantheman720 I had one: Welcome to the SwitchYard Quickstart Repository which is going to the wrong spot, but it seems like we don't really want that one anyway. It's being placed at /quickstarts/fues/switchyard/index.html seems like it's a root README which simply doesn't match our requirements (it doesn't have a topic for example) because it isn't actually a quickstart. @paulrobinson Not sure what we want to do with that. I thought We wanted to just not include those that didn't fit the criteria.

LightGuard:
retest this please

LightGuard:
retest this please

PR #1080, opened 2016-05-05
base/Dockerfile Red Hat GPG key location URL update

rctrh:
* curl of /etc/pki/rpm-gpg/RPM-GPG-KEY-redhat-release was failing
  silently due to a 301. curl doesn't follow the rediect by default.
* Changed URL from https://www.redhat.com/security/data/fd431d51.txt
  to https://www.redhat.com/security/data/fd431d51.txt


LightGuard:
this is ok to test

PR #1082, opened 2016-05-06
Tests should handle hosts with a trailing forward slash

RubyGemTSL:


RubyGemTSL:
@paulrobinson - could you please merge this asap? It's causing failures on the production smoke tests.

RubyGemTSL:
also to the nightly tests and staging build.

PR #1087, opened 2016-05-10
Updated product version and links for 6.6

huffmanca:
Need to update the developers.redhat.com documentation links for JDG 6.6.


paulrobinson:
this is ok to test

PR #1088, opened 2016-05-10
Fixing issue with dcp

RubyGemTSL:


LightGuard:
I've come to the same conclusion.

PR #1096, opened 2016-05-12
Fixes DEVELOPER-2947 (fuse connector page)

LightGuard:
Please see https://issues.jboss.org/browse/DEVELOPER-2947 for details.

In Drupal it was 'fixing' bad HTML. Well, turns out it thought the HTML
in the script tags was bad (probably because of our template marker
choice), so it did it's best to fix it. Turns out it actually broke it.
I've created a new html text format that only administrators can use
which doesn't do this correction. I named it 'as-is HTML' so Drupal
doesn't try to fix anything.

In order to test this you'll have to start up Drupal:

```
cd _docker
bundle exec ./control.rb -u --run-the-stack
```

Once that finishes you'll need to visit the fuse connectors page to
verify things are working correctly.

### DIFF
--- a/_docker/drupal/modules/custom/redhat_developers/config/install/filter.format.as_is_html.yml
+++ b/_docker/drupal/modules/custom/redhat_developers/config/install/filter.format.as_is_html.yml
@@ -1,0 +1,28 @@
+uuid: 5610f355-4f75-4ee8-a600-b605c9f99a53
+langcode: en
+status: true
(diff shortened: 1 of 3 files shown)


LightGuard:
retest this please

Dantheman720:
+1 Looks good to me!

LightGuard:
@RubyGemTSL mind merging this one too, please?

PR #1101, opened 2016-05-13
I think I have something that works

LightGuard:
Putting this up here so I can test in the staging environment.


LightGuard:
retest this please

LightGuard:
Well, it looks like it finally worked. @RubyGemTSL @robpblake ready to merge I believe.

_docker/control.rb
```diff
@@ -33,23 +33,6 @@ def modify_env
   end
 end
 
-def set_ports
```
robpblake: It is a good simplification that we're getting rid of this!

_docker/control.rb
```diff
@@ -177,22 +160,25 @@ def block_wait_searchisko_started(supporting_services)
...
+    network_exists = false
   end
 
-  if tasks[:set_ports]
-    puts 'Setting ports...'
-    set_ports
-    # Output the new docker-compose file with the modified ports
-    File.delete('docker-compose.yml') if File.exists?('docker-compose.yml')
-    File.write('docker-compose.yml', ERB.new(File.read('docker-compose.yml.erb')).result)
+  if tasks[:kill_all] && File.exists?('docker-compose.yml') && network_exists
```
robpblake: I think ideally we should unit test logic like this by pulling it into a function e.g something like:

`
if tasks[:kill_all]
 kill_current_environment(tasks)
end 
`

And then the kill_current_environment task can be unit tested.

If not that, then I think we should at least document the logic behind this with a comment e.g. we only run docker-compose down if there is a network because the network could only exist if docker-compose up has run previously.

robpblake:
From what I can see, this *should* be fine. I guess we'll only really know once it gets merged.

@LightGuard Just a few minor comments for you to respond to.

cheers,

Rob

LightGuard:
retest this please

robpblake:
@LightGuard - Thanks for adding the tests.

@paulrobinson / @RubyGemTSL From my perspective this looks good to merge. Can you review and merge please.

cheers,

Rob

PR #1106, opened 2016-05-16
Fixes DEVELOPER-3023

LightGuard:
See https://issues.jboss.org/browse/DEVELOPER-3023 for full info.

I've changed this a bit, turns out we don't need the images to be in
Drupal after all. As long as they're in the web root for the
application, Apache will just serve them up as normal. We don't have to
worry about adding the images/fonts and creating nodes. What we're doing
now is mounting them from the host.

I had to fix the Gulp script a bit so it doesn't do all of the changes
for CDN. I also removed any CDN info for Drupal from the site.yml. I
removed the code to create all the nodes for images / fonts inside
Drupal when the redhat_developers module is installed. Also had to
change the chrome inside Drupal to reference the images in the correct
places.

Lastly, I made one small change to take out the left side floating links
for Drupal when you're logged into Drupal We could change / add back in
these things at a later time if we decide we need to.


LightGuard:
@robpblake please take a look at this and let me know, or have Paul merge it if you're happy with it.

robpblake:
@LightGuard Looks good! Really nice simplification!

@paulrobinson Can you just give this the once over and merge if you're happy.

cheers,

Rob

PR #1108, opened 2016-05-17
[DEVELOPER-2936] Updated Gemfile to use HTTPS to fetch Git dependencies

robpblake:
Whilst trying to get things to build on the new VMs for the Drupal environment, a persistent stumbling block was the fact that a couple of dependencies are pulled in using the git:// protocol.

The problem occurs on the VMs in that they are behind a proxy and I could not get bundle install to work behind the proxy.

As soon as I moved fetching the dependencies to using HTTPs, then everything works absolutely fine. It would also appear that HTTPS is the recommended approach for Git dependencies as-per the "Security" section on this page: http://bundler.io/git.html

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -17,7 +17,7 @@ end
 
 # GEMS
 #gem 'awestruct', '0.5.7'
-gem 'awestruct', github: 'lightguard/awestruct', branch: 'feature/perf-testing-large-site'
+gem 'awestruct', git: 'https://github.com/lightguard/awestruct', branch: 'feature/perf-testing-large-site'
 #gem 'awestruct', path: '~/projects/ruby/awestruct'
 gem 'slim', '~> 3.0'
 gem 'kramdown', '~> 1.0.1'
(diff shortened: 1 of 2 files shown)


robpblake:
@LightGuard - Could you please review this as I'm not 100% this the best way to update the dependencies to use HTTPS (my knowledge of Gemfile and Gemfile.lock is very lacking)

LightGuard:
Yep, this should be fine.

On Tuesday, May 17, 2016, Rob Blake <notifications@github.com> wrote:

> @LightGuard <https://github.com/LightGuard> - Could you please review
> this as I'm not 100% this the best way to update the dependencies to use
> HTTPS (my knowledge of Gemfile and Gemfile.lock is very lacking)
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/redhat-developer/developers.redhat.com/pull/1108#issuecomment-219724659>
>


-- 
Sent from Gmail Mobile

PR #1110, opened 2016-05-17
[DEVELOPER-3041] Do not expose Awestruct port on Jenkins

robpblake:
Another attempt to resolve the problem with "port in use" errors. This time we only expose the awestruct port if we're running in a local development environment i.e. not on Jenkins.

### DIFF
--- a/_docker/docker-compose.yml.erb
+++ b/_docker/docker-compose.yml.erb
@@ -62,10 +62,7 @@ services:
     build: ./awestruct
     command:
       - "rake git_setup clean preview[docker]"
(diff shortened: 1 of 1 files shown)


robpblake:
@LightGuard Any reason why this would cause us a problem? I've simply stopped binding the Awestruct port to the host unless you are running in a dev environment.

paulrobinson:
I should have mentioned this before, but it was confusing me as to why we were exposing 4242 in anything but dev. 

I believe this port is only used when Awestruct acts as a web server. This happens in dev, but in all other environments it simply generates static content which is rsynced to a separate Apache server.

PR #1113, opened 2016-05-17
DEVELOPER-3043: Missing node.js tab on rhel get started

rctrh:
* Add missing Node.js jab to Step 4.
* Fix incorrect links in Step 4.
* Clean up step 4 TTHW for consistency and accuracy.
* Re-org FAQ into installation and development.
* make better use of RHdev content in FAQ


LightGuard:
this is ok to test

PR #1116, opened 2016-05-18
[DEVELOPER-2955] Implementation to take a backup of drupal environment

robpblake:
This is the first implementation of backup of drupal environments.

It is worth reading the accompanying README.md

 The main points of interest are:

- _docker/lib/backup/backup.rb -> Runs the backup
- _docker/lib/backup/git_backup_strategy.rb -> A strategy for storing backups in Git

Everything has been unit tested and control.rb has been updated to support running backups.

syntax for running a backup is:
```
./control.rb --backup
```
or for a named backup

```
./control.rb --backup my-backup-name
```

The backup will only run in the production environment.


robpblake:
@LightGuard / @paulrobinson This is the implementation for the backup of Drupal environments.

This implementation draws together a lot of the concepts that we've been discussing over the past few weeks. Can you please both review and make sure that the approach is clear and is in-line with your expectations. I suggest starting with the accompanying README.md. As much as possible, I've taken the approach of testing everything which should give us confidence that things are working as we want them to.

@LightGuard I would appreciate a code review on the Ruby. I am sure it is not very idiomatic, but I'm coming at it from 10+ years of Java development so it will take me sometime to pick-up the Ruby best practices.

cheers,

Rob

Rakefile
```diff
@@ -51,7 +51,7 @@ Rake::TestTask.new do |t|
   t.libs = ["_docker/lib"]
   t.warning = false
   t.verbose = true
-  t.test_files = FileList['_docker/test/*.rb', '_tests/*.rb'] #Let's add more files here!
+  t.test_files = FileList['_docker/test/**/*.rb', '_tests/*.rb'] #Let's add more files here!
```
LightGuard: I've wondered why we named _docker/test instead of _docker/tests but I guess it really doesn't matter.

_docker/docker-compose.yml.erb
```diff
@@ -154,7 +154,23 @@ services:
       - DB_PASSWORD=searchisko
     links:
      - mysql
-  
+
+  #
+  # Environment actions
+  #
+  backup:
+    build: ./awestruct
```
LightGuard: Any particular reason to use the awestruct container vs the base ruby container for stuff like this?

_docker/lib/backup/backup.rb
```diff
@@ -0,0 +1,112 @@
...
+#
+# The purpose of this class is to perform a backup of a RHD environment.
+# It can be passed a single argument, which will name the folder into which the backup is placed.
+#
+# @author rblake@redhat.com
+#
+class Backup
+
+  # Default name of the Drupal filesystem backup
+  @@DRUPAL_FILESYSTEM_TAR = 'drupal-filesystem.tar.gz'
```
LightGuard: Is this class used multiple times? I don't see a reason to use class variables for this versus instance variables

_docker/lib/backup/backup.rb
```diff
@@ -0,0 +1,112 @@
...
+  def self.drupal_database_sql_dump
+    @@DRUPAL_DATABASE_SQL_DUMP
+  end
+
+  @backup_directory
+  @backup_strategy
+  @drupal_directory
+  @database_name
+  @process_runner
+  @log
```
LightGuard: You don't need to declare these outside of the initialize method

_docker/lib/backup/backup.rb
```diff
@@ -0,0 +1,112 @@
...
+  # Generates a user-friendly timestamp for the back-up
+  #
+  private def generate_backup_time
+    DateTime.now.strftime('%F-%H-%M-%S')
+  end
+
+  def execute(args)
+
+    backup_time = generate_backup_time
+    tag = args.first.to_s == '' ? backup_time : args.first
```
LightGuard: Looks like you're assuming args is going to be an array. You may want to specify it as an array up in the method declaration (args = []) just in case, and an args.nil? check isn't a bad idea either. args.first.to_s could become args.first.empty?

_docker/lib/backup/git_backup_strategy.rb
```diff
@@ -0,0 +1,47 @@
...
+#
+# A backup strategy that pushes files to a provided Git repository
+#
+# @author rblake@redhat.com
+#
+class GitBackupStrategy
+
+  @log
+  @process_runner
+  @git_repo_location
```
LightGuard: These don't need to be specified outside of the initialize method either.

_docker/lib/process_runner.rb
```diff
@@ -0,0 +1,73 @@
...
+# @author rblake@redhat.com
+#
+class ProcessRunner
+
+  class ProcessFailedError < StandardError
+
+  end
+
+  @verbose
+  @log
```
LightGuard: Same here

_docker/lib/process_runner.rb
```diff
@@ -0,0 +1,73 @@
...
+  # <p>Execute the given command specified by cmd.</p>
+  #
+  # @throws - ProcessFailedError if the command exits with a non-zero status
+  #
+  def execute!(cmd)
+
+    cmd = prepare_command_for_execution(cmd)
+    @log.info("Executing command '#{cmd}'")
+
+    Open3.popen3(cmd) { |_, stdout, stderr, wait_thr|
```
LightGuard: For larger blocks the do ... end syntax is prefered over the curly brackets.

_docker/lib/process_runner.rb
```diff
@@ -0,0 +1,73 @@
...
+          while ( ! i.eof? )
+            @log.info("(Console): #{i.readline}")
+          end
+        end
+        threads << Thread.new(stderr) do |i|
+          while ( ! i.eof? )
+            @log.error("(Console): #{i.readline}")
+          end
+        end
+      end
```
LightGuard: There's no threads.each {|t| t.join} so we're never waiting for them to finish?

_docker/lib/process_runner.rb
```diff
@@ -0,0 +1,73 @@
...
+
+      exit_code = wait_thr.value
+      determine_process_status(cmd, exit_code)
+    }
+  end
+
+  private :determine_process_status, :prepare_command_for_execution
```
LightGuard: Do you plan on using this syntax in the rest of the files for consistency (this is how I've done it)?

_docker/test/backup/test_backup.rb
```diff
@@ -0,0 +1,49 @@
...
+require_relative '../../../_docker/lib/backup/backup'
+
+class TestBackup < MiniTest::Test
+
+  @backup_directory
+  @database_name
+  @drupal_filesystem
+  @git_backup_strategy
+  @process_runner
+  @backup
```
LightGuard: Same here

LightGuard:
Save the comments I've made this looks good. Do you have examples for the mysql config files? How about which git repo, or are you just testing using your own account?

_docker/lib/process_runner.rb
robpblake: I actually started out using this syntax and then reverted to putting the private declaration on the method. I found I was completely forgetting to do it with using the symbol approach. This is again going to be a hangover from my Java experience, so I'm happy to go with whichever approach is considered best practice in Ruby-land. Any preference?

_docker/lib/process_runner.rb
robpblake: Good spot.

Rakefile
robpblake: Re-named to tests for consistency.

_docker/docker-compose.yml.erb
robpblake: Primarily due to the way in which we're handling Gemfiles at the minute. control.rb is copying in Gemfile and Gemfile.lock from the project root to keep things "in-sync" when the awestruct image gets built. It just felt easier to adopt this for now.

What I'd like to do at a later date is pull all of these actions e.g. backup, restore, export out into their own Git repository and distribute them as an image. When we do that then we can:

- Have our own Gemfile and Gemfile.lock for just that project
- Distribute them on a Docker image that only has the bare-minimum. I actually like the idea of using https://github.com/docker-library/ruby/blob/5d04363db6f7ae316ef7056063f020557db828e1/2.1/onbuild/Dockerfile as it makes it very easy to build Docker images that are basically ruby distributions.

So again its a bit of a "definitely room for improvement" type of scenario, which I think we can look at post v1.

robpblake:
I'll add an example my.cnf.

In terms of the Git repo, it can work with any. As long as the credentials are passing to container and runtime and the Git repository is mounted at /backups inside the container, it really doesn't care. I tested it out with one of my private repos, but we should use a private repo under the Red Hat banner for when we go live.

The setup of that repository is covered under user story https://issues.jboss.org/browse/DEVELOPER-2955.

_docker/docker-compose.yml.erb
LightGuard: Okay, I'm good with leaving it for now and we can look at it again post v1.

robpblake:
@paulrobinson / @LightGuard Nothing further from me on this. If you're both happy then this can be merged.

cheers,

Rob

robpblake:
@LightGuard Can you please merge this if you're happy. Paul has no further comments, but wanted to ensure that all your questions have been answered.

PR #1117, opened 2016-05-19
Made description unique

paulrobinson:
This is simple change to prevent two pages having the same description. I'm checking that making this fix percolates through the DCP instances and improves the site search.


paulrobinson:
@RubyGemTSL can you review please?

PR #1120, opened 2016-05-23
[DEVELOPER-3024] Support for running Drupal in multiple environments

robpblake:
**ONLY MERGE THIS AFTER #1116** 

This pull request updates the control.rb script to add support for different environments.

The environment that you now want to use is specified by using the -e parameter on the command line. For example:

`bundle exec ./control.rb -e awestruct-dev -t`

Reading the environments README.md file is highly recommended for an explanation for the environments supported.

The default environment for the script is the Awestruct Pull request environment. The reason for this is that it means that we should be able to merge this without making any changes to the current pull request job on Jenkins.

Where possible everything has been unit tested.

@LightGuard / @paulrobinson I'd appreciate an overall review and your comments please.

@RubyGemTSL FYI: I've split out all test docker definitions into their own docker-compose.yml file. You'll find an environment called `testing` in the environments directory. This environment is merged into all other environments at runtime, meaning that you only need to define test dependencies once.

**ONLY MERGE THIS AFTER #1116**


robpblake:
retest this please

robpblake:
Can't get Searchisko to start on pull request build..just hangs.

robpblake:
retest this please

robpblake:
@LightGuard I've not put any values into rhd.settings.php for the drupal-dev and drupal-pull-request environments as yet (other than to test that templating works as expected in the drupal-pull-request environment) because I'll probably need your help to work out exactly what we need in there.

You should also be able to integrate your new image into this fairly easily. Just update the environment docker-compose.yml where Drupal is used with the build for the new image (or if the build is still in _docker/drupal, then this should just work).

cheers,

Rob

PR #1123, opened 2016-05-24
[DEVELOPER-3022] Deploy Awestruct content changes into Drupal

robpblake:
This pull request introduces the following changes:

- Creates an Awestruct profile for each of the Drupal environments
- Moves the default Awestruct command to be run into the docker-compose.yml file for each environment
- Removes support for the --drupal-nightly option in control.rb
- Correctly sets the Drupal and Searchisko ports in drupal-dev environment to support hard-coded endpoints in twig files
- Updates to options.rb to support the above

### DIFF
--- a/_config/site.yml
+++ b/_config/site.yml
@@ -80,10 +80,52 @@ profiles:
 <% end %>
     push_to_searchisko: true
 
-  drupal:
+  drupal_dev:
     <<: *docker
-    base_url: http://<%= ENV['DRUPAL_HOST_IP'] || 'docker' %>:<%= ENV['DRUPAL_HOST_PORT'] %>/
+    base_url: http://<%= ENV['DRUPAL_HOST_IP']%>/
+    dcp_base_protocol_relative_url: //docker:<%= ENV['SEARCHISKO_HOST_PORT'] %>/
     drupal_base_url: http://drupal/
+
+  drupal_pull_request:
+    <<: *docker
+    base_url: http://<%= ENV['ACCESSIBLE_SLAVE_IP']%>:<%= ENV['DRUPAL_HOST_PORT'] %>/
+    dcp_base_protocol_relative_url: //<%= ENV['ACCESSIBLE_SLAVE_IP'] %>:<%= ENV['SEARCHISKO_HOST_PORT'] %>/
+    drupal_base_url: http://drupal/
+
+  drupal_staging:
+    <<: *profile
+    base_url: http://developer-drupal.web.stage.ext.phx2.redhat.com/
+    drupal_base_url: http://drupal/
+    download_manager_file_base_url: //developers.stage.redhat.com/download-manager/file/
+    download_manager_base_url: https://developers.stage.redhat.com
+    dcp_base_url: http://dcp.stage.jboss.org/
+    dcp_base_protocol_relative_url: //dcp.stage.jboss.org/
+    keycloak_account_url: https://developers.stage.redhat.com/auth/realms/rhd/account/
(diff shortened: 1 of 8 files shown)


robpblake:
@LightGuard I've created a profile per Drupal environment in _config/site.yml. For the Drupal staging and production environments I've pretty much copied across the current staging and production configurations.

Could you please give this the once over to make sure it makes sense.

cheers,

Rob

_config/site.yml
```diff
@@ -80,11 +80,53 @@ profiles:
 <% end %>
     push_to_searchisko: true
 
-  drupal:
+  drupal_dev:
     <<: *docker
-    base_url: http://<%= ENV['DRUPAL_HOST_IP'] || 'docker' %>:<%= ENV['DRUPAL_HOST_PORT'] %>/
+    base_url: http://<%= ENV['DRUPAL_HOST_IP']%>:<%= ENV['DRUPAL_HOST_PORT'] %>/
```
LightGuard: Any reason we can't hard code this now? Simply `docker` or `localhost`

_config/site.yml
```diff
@@ -80,11 +80,53 @@ profiles:
...
+    client_side_error_logging: true
+    build_threads: 1
+    dtm_code: //assets.adobedtm.com/2b327a0a54b320bf1bcdb5fa39a2b896027067d9/satelliteLib-9ec3c34e9d660fa2fcb85ced512c3a48dc09c0c5-staging.js
+    deploy:
+      <<: *deploy
+      path: /stg_htdocs/it-rhd-stg-main
+
+  drupal_production:
+    <<: *profile
+    base_url: http://developer-drupal.web.prod.ext.phx2.redhat.com/
```
LightGuard: This should certainly be developers.redhat.com

_config/site.yml
```diff
@@ -80,11 +80,53 @@ profiles:
...
 
+  drupal_pull_request:
+    <<: *docker
+    base_url: http://<%= ENV['ACCESSIBLE_SLAVE_IP']%>:<%= ENV['DRUPAL_HOST_PORT'] %>/
+    dcp_base_protocol_relative_url: //<%= ENV['ACCESSIBLE_SLAVE_IP'] %>:<%= ENV['SEARCHISKO_HOST_PORT'] %>/
+    drupal_base_url: http://drupal/
+
+  drupal_staging:
+    <<: *profile
+    base_url: http://developer-drupal.web.stage.ext.phx2.redhat.com/
```
LightGuard: I don't think this is the correct url for staging.

LightGuard:
Few comments for you @robpblake

_config/site.yml
robpblake: As per our discussion last night, I think developer-drupal.web.prod.ext.phx2.redhat.com is actually correct. This will mean that all URLs point to Drupal until the content is exported by wget. 

We will have to ensure that wget runs with the --convert-links option to make all links on the page relative to one another.

_config/site.yml
robpblake: See comment above.

_config/site.yml
robpblake: "localhost" won't work for those who have to run Docker through docker-machine i.e anyone with a mac.

control.rb now attempts to take into account whether you're using docker-machine or not. If yes, it expects a host alias for docker to be set, if not then it assumes you're running docker on localhost. In either case, control.rb sets the value of ENV['DRUPAL_HOST_IP'] correctly, so there is no need for us to use the "docker" alias here.

PR #1125, opened 2016-05-25
[DEVELOPER-3081] Ensured that we invoke npm install during Drupal builds

robpblake:
Small update to control.rb to ensure that we invoke 'npm install' if we're running in a Drupal-based environment.


robpblake:
@paulrobinson / @LightGuard Just a small update to control.rb to ensure that we run 'npm install' if we're running in a Drupal-based environment. 

Appreciate a review and merge please.

PR #1129, opened 2016-05-26
docker env determine_docker_host fix

rctrh:
control.rb::determine_docker_host_for_container_ports was trying to
get the IP address of the system hostname, which might not be defined on
developer's machines, before checking whether the hostname `docker`
would resolve.

The expectation is developer's machines with an RHDev docker
environment will have docker defined as a hostname with an
ip address for localhost.

This workaround will still fail cryptically if the hostname docker isn't
defined and the current hostname doesn't resolve to an ip address.

### DIFF
--- a/_docker/control.rb
+++ b/_docker/control.rb
@@ -247,12 +247,11 @@ end
 #
 def determine_docker_host_for_container_ports
 
-  docker_host = Resolv.getaddress(Socket.gethostname)
-
   begin
     docker_host = Resolv.getaddress('docker')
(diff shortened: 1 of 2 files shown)


LightGuard:
this is okay to test

LightGuard:
this is ok to test

LightGuard:
@robpblake Mind taking a look at this as well?

rctrh:
any chance this (or a replacement) will get merged soon  (so I don't have to keep re-applying when switching branches?)

PR #1134, opened 2016-05-31
Added option to override title on product pages.

unibrew:
This pull request allows overriding page title on pages inheriting from product-base.html.slim layout.

### DIFF
--- a/_layouts/product-base.html.slim
+++ b/_layouts/product-base.html.slim
@@ -5,7 +5,7 @@ secondary_section: supported-products
 ---
 /The base layout for products, used to provide a custom left bar or right bar
 /If you are aren't sure which product layout to use, you probably want 'product', not this one!
-- page.title = "#{page.product.abbreviated_name} - #{page.type}"
+- page.title = page.title ? page.title : "#{page.product.abbreviated_name} - #{page.type}"
 - page.product.highlight_colour = '#8FD102' unless page.product.highlight_colour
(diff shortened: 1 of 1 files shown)


paulrobinson:
this is ok to test

paulrobinson:
Do you have an example of where I can see this working?

unibrew:
I'm not yet on VPN to double check but difference in title should be visible between those two pages:
http://developers.redhat.com/products/softwarecollections/get-started-rhel7-nodejsbeta/

http://developers-pr.stage.redhat.com/pr/1134/build/1288/products/softwarecollections/get-started-rhel7-nodejsbeta/

unibrew:
doublechecked, the second link contains the correct, longer title

rctrh:
@unibrew Thanks that takes care of the title.  However the :awestruct-description: isn't being picked up from the file.  Should this fix have address that as well?

unibrew:
I didn't spot that description doesn't work. I'm currently fixing descriptions in content and I already noticed some that are problematic. So, I'll provide a separate fix for this and let's leave this PR to be only about titles.

rct:
Ok, in that case, this looks good.  When do you you anticipate merging this?

unibrew:
I cannot merge it because I'm not part of the frontend team, just helping out temporarily.
@paulrobinson was planning to merge it today.

PR #1139, opened 2016-06-01
[DEVELOPER-308] Ensure docker images can build behind a proxy

robpblake:
This pull request allows for all images used by the project to be built in environments where a proxy server is required for Internet access. This primarily relates to the drupal staging and drupal production environments.

Changes in this pull request:

1. Added a build ARG to images that need to be built in environments with proxies that allows specification of proxy server at runtime
2. Bumped version of core project docker images to 2.0.0 to avoid updating the images used by other pull requests until this is merged
3. Removed the `unit_test` service definition in docker-compose.yml and instead made use of the awestruct image defined in each environment.
4. Move the `acceptance_test` service definition into the docker-compose.yml for environments in which they are supported i.e. not drupal-staging and drupal-production
5. Updated to control.rb and supporting code to supply build-time args for images if we are running in an environment that needs a proxy server
6. Unit tests for the above


robpblake:
@LightGuard Could you please take a look at this and review the changes.

I'd also appreciate it if you could just '--run-the-stack' in awestruct-dev on your machine just to make sure things are working for others.

cheers,

Rob

LightGuard:
@robpblake would you want to incorporate https://github.com/lightguard/developers.redhat.com/tree/fetaure/slimming_docker into this?

PR #1147, opened 2016-06-02
[DEVELOPER-2937] No supporting services for backup

robpblake:
When taking a backup, we do not need to start or wait for any supporting services.


robpblake:
@LightGuard If you could review and merge I would be grateful.

cheers,

Rob

PR #1156, opened 2016-06-08
[DEVELOPER-3134] Reduced the number of chrome nodes to 2 by default

robpblake:
To try and lower resource contention on stumpjumper, reduce the number of chrome nodes to 2 per pull request.

### DIFF
--- a/_docker/lib/options.rb
+++ b/_docker/lib/options.rb
@@ -76,7 +76,7 @@ class Options
         end
 
         if ENV['RHD_BROWSER_SCALE'].to_s.empty?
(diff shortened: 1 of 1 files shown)


robpblake:
@RubyGemTSL I've reduced the number of chrome nodes to 2 and this increases the test time to about 2 minutes from 1 minute 15 seconds. I think that extra 45 seconds is a reasonable trade-off to lower the amount of resources consumed by each PR?

cheers,

Rob

RubyGemTSL:
yep - looks good to me. Will merge when Blinkr is finished.

PR #1157, opened 2016-06-08
[DEVELOPER-3080] Removed html based redirects on products index pages.

unibrew:
Related to comments in the task.


paulrobinson:
this is ok to test

PR #1163, opened 2016-06-09
Allowing asciidoctor includes for sections

LightGuard:
Sections were not working with includes before, they are now.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,6 +1,6 @@
 GIT
   remote: https://github.com/awestruct/aweplug
-  revision: 23a8c06f42e5b73cfe122fffd0023b2975cadf72
+  revision: c89287eddc88d30ca6e1b65309c26b0392166629
   specs:
     aweplug (1.0.0.a24)
       curb (~> 0.8.5)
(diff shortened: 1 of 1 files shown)


LightGuard:
retest this please

robpblake:
retest this please

RubyGemTSL:
retest this please

RubyGemTSL:
retest this please

RubyGemTSL:
retest this please

LightGuard:
@rctrh has verified this works for him.

One small caveat is that anything you're including inside sections (which are being rearranged via a layout) must have a lower section heading (if the section being included has a level 2 heading the headings in the included content must be level 3 or lower) level otherwise things break without any apparent cause.

robpblake:
I suspect @adelasofia is better placed than I am to say if this works or not.

cheers,

Rob

LightGuard:
@adelasofia You'll need to try an asciidoctor include in a page that uses sections to verify this works. When you do the include make sure you're make it relative to the base of the project, not the file you're adding the include to.

PR #1167, opened 2016-06-13
gitignore emacs backup files *~

rctrh:
Because some luddites, like me, still use emacs.


RubyGemTSL:
ok to test

RubyGemTSL:
this is ok to test

PR #1170, opened 2016-06-14
[DEVELOPER-3174] Prevent intermittent unit test failure due to env variable clash.

robpblake:
Ensured that we revert the value of the cdn_prefix back to what it originally was after each test run.


robpblake:
@paulrobinson / @RubyGemTSL / @LightGuard A small fix to the unit tests to avoid intermittent failures. Can one of you please review and merge if you're happy.

cheers,

Rob

PR #1180, opened 2016-06-16
[DEVELOPER-3200] Fix intermittent test failure

robpblake:
Ensure that we set env variables to nil after the test runs to prevent interference with subsequent tests.


robpblake:
@RubyGemTSL / @LightGuard A small change to fix another test that intermittently fails due to environment variables clashing.

cheers,

Rob

PR #1181, opened 2016-06-16
DEVELOPER-2938 Docs for Drupal restore

LightGuard:
See https://issues.jboss.org/browse/DEVELOPER-2938 for more information.

In order to properly test this you'll need to get a backup. Currently
you can use my backups located at
https://github.com/LightGuard/rhd-backup. There should be three backups
in that repository, they're in tags. You can use the most current backup
from 2016-06-15 to test, though any of them will work.


_docker/drupal/README.md
```diff
@@ -35,6 +35,19 @@ The `composer create-project` command passes ownership of all files to the
 project that is created. You should create a new git repository, and commit 
 all files not excluded by the .gitignore file.
 
+## Restoring content to Drupal
+
+If something goes wrong and you need to restore the drupal environment from a backup the steps are pretty easy:
+
+1. Obtain the backup. This will include an sql file an a tar.gz file.
+2. Extract the tar.gz file into `_docker/drupal/web`.
+3. Execute `control.rb -e drupal-dev --run-the-stack`.
```
robpblake: @LightGuard Perhaps make it clear that the person following these instructions needs to specify the environment (and also container names) correctly. For example, these instructions won't work if I try them in production environment as the container names will be different

robpblake:
@LightGuard Bar my comments on making it clearer in the instructions that you need to correctly specify the environment, I'm happy for this to be merged.

Also can I just double check that this will work where the backup.tar.gz only contains the `sites/default/images` directory? When we spoke with Phase2, `sites/default/images` was the only directory they said that needed to be backed up when using the composer build approach. Given that all Drupal configuration is in Git, my assumption is that to do a restore in production we would:

1. Stop all current containers
2. Checkout the Git repository to a known working state
3. `--run-the-stack`
4. Restore the database as per your instructions
5. Restore `sites/default/images` from the backup

Have I understood how it works correctly?

Dantheman720:
@LightGuard @robpblake - I think Step 5. in the documentation should say "Obtain a shell for the drupalmysql docker container using: docker exec -it drupaldev_drupalmysql_1 /bin/bash"

The current docker exec is pointing to drupaldev_drupal_1.

_docker/drupal/README.md
```diff
@@ -35,6 +35,19 @@ The `composer create-project` command passes ownership of all files to the
...
+4. Copy the sql file into the drupalmysql container using `docker cp /path/to/drupal-db.sql drupaldev_drupalmysql_1:/drupal-db.sql`
+5. Obtain a shell for the drupalmysql docker container using: `docker exec -it drupaldev_drupal_1 /bin/bash`
```
Dantheman720: This should be drupaldev_drupalmysql_1 not drupaldev_drupal_1

robpblake: @Dantheman720 Yes you're right, well spotted.

LightGuard:
@Dantheman720 @robpblake Got that instance for the mysql container fixed up. Anything else here or are we good to go (assuming tests pass)?

robpblake:
@LightGuard Still a couple of comments from me.

1. The instructions are tied to the drupal-dev environment. Perhaps just make people aware that they need to change the environment (and container names) if they want to run this in other environments. The environment in which we're most likely to run this is production

2. See my previous comment ^ about whether this will work if we just have the `sites/default/images` directory within the tarfile.

LightGuard:
Yes, it should work fine if we just restore the site/default/images directory.

LightGuard:
@robpblake Will that change work?

robpblake:
@LightGuard Can you merge master into your branch. There are currently conflicts.

cheers,

Rob

LightGuard:
Weird, I didn't have any merge conflicts locally.

LightGuard:
@robpblake Please look it over and merge if you're happy with it now.

PR #1193, opened 2016-06-21
[DEVELOPER-3202] Remove bind mount of Drupal filesystem

robpblake:
This pull request changes the way in which we build the Drupal Docker image so that we do not have to bind mount the entire Drupal filesystem into the running container.

The driver for this change was my work attempting to get pull requests to build using Drupal. With the Drupal filesystem bind mounted into the container, I ran into numerous permissions issues and also Jenkins was unable to clean the workspace due to folders being in use.

In this pull request, I have:

1. Updated the Drupal Dockerfile to `COPY` in the required parts of the filesystem to the container image
2. Re-structured the `_docker/drupal` directory to logically group together filesystem components to aid with Docker caching when building the Drupal image
2. Split templating of resources in the `drupal-pull-request` environment into 2 parts. First we create the resources so that they can be bind mounted into the container, and then we template the contents after the containers are running.


robpblake:
@LightGuard Could you please test this PR locally. After running:

`control.rb -e drupal-dev --run-the-stack`, you should get the site in Drupal and then a `git status` should be clean showing no modifications.

cheers,

Rob

LightGuard:
Very happy with this.

PR #1217, opened 2016-06-26
DEVELOPER-3210 Drupal production build

LightGuard:
See https://issues.jboss.org/browse/DEVELOPER-3210 for full information.

These are changes that need to happen for the production build to work correctly. The jenkins job (developers.redhat.com-drupal) should be working, however, we're talking to the production dcp currently and I haven't done a complete test, but what I have tested has worked thus far.


paulrobinson:
FYI, the acceptance tests will fail on the tests for the /products and /forums page. I needed to merge the product page changes after we spent too long trying to fix the tests. Hopefully @RubyGemTSL can help us out first thing on Monday.

RubyGemTSL:
tests will fail due to failing tests being pushed over the weekend. I will have a PR merged soon to fix this.

LightGuard:
There may be more to this branch.

RubyGemTSL:
retest this please

LightGuard:
@robpblake Take a look. I've verified running this in a different box (my local box in this case) to the prod drupal instance works fine. We'll just need to create the jobs and set them all up.

robpblake:
retest this please

_docker/environments/drupal-production/docker-compose.yml
```diff
@@ -43,9 +43,9 @@ services:
       - "80:80"
     volumes:
       - /credentials/drupal/rhd.settings.yml:/var/www/drupal/web/sites/default/rhd.settings.yml
-      - ./rhd.settings.php:/var/www/drupal/web/sites/default/rhd.settings.php
-      - ../../../images:/var/www/html/images:ro
-      - ../../../stylesheets/fonts:/var/www/html/fonts:ro
+      - /credentials/drupal/rhd.settings.php:/var/www/drupal/web/sites/default/rhd.settings.php
+      - ../../../images:/var/www/drupal/web/images:ro
+      - ../../../stylesheets/fonts:/var/www/drupal/web/fonts:ro
```
robpblake: We need to ensure that we have the bind-mount for the `/var/www/drupal/web/sites/default/images` directory so that images uploaded to Drupal are stored on the NAS disk attached to the VM.

Perhaps something like this to make it extra clear what purpose the bind mount serves?

```
volumes:
 - /data/drupal-images:/var/www/drupal/web/sites/default/images
```

What do you think?

_docker/environments/awestruct-production/docker-compose.yml
```diff
@@ -0,0 +1,36 @@
...
+  #
+  # Testing
+  #
+  unit_tests:
+    build:
+      context: ../../awestruct
+    volumes:
+      - ../../../:/home/awestruct/developer.redhat.com
+    entrypoint: "bundle exec rake test"
+
```
robpblake: Just so I'm following, the `awestruct-production` environment is the one that runs on Stumpjumper to push content into Drupal?

_docker/environments/drupal-production/rhd.settings.php
```diff
@@ -6,3 +6,6 @@
   $config['redhat_developers'] = $yml_settings;
 }
 
+$settings['hash_salt'] = 'xuAWpK0fmrZ6UGofFcP3lBkcmdpdumWMLqvCbnYjFY85OgRXYvEKPItJDH66vs4UpeYORQXLHQ';
+$settings['install_profile'] = 'standard';
+$config_directories['sync'] = 'sites/default/files/config_BbPlfGDu86LJqlRwhA9RbCf38VZMDijNF-owvfhuzVL73hk7BtWwy3kfIlqKLXeiSgTA-MHeVw/sync';
```
robpblake: Do we need this file? In the docker-compose.yml for `drupal-production` I see the following in the volumes definition for Drupal:

```
- /credentials/drupal/rhd.settings.php:/var/www/drupal/web/sites/default/rhd.settings.php
```

Which would suggest this file is not being used?

robpblake:
@LightGuard A few comments from me.

cheers,

Rob

_docker/environments/drupal-production/rhd.settings.php
LightGuard: No, we no longer need it, good catch

_docker/environments/awestruct-production/docker-compose.yml
LightGuard: Correct

LightGuard: I thought about naming it awestruct-drupal-production, but then it has drupal in it and it would start the drupal environment

_docker/environments/drupal-production/docker-compose.yml
LightGuard: Yep, sorry, I'll add it.

_docker/environments/awestruct-production/docker-compose.yml
robpblake: Rather than creating another environment, could we just use the existing drupal-production one and simply remove the `links:` directive from the `awestruct` service definition in the `drupal-production/docker-compose.yml`?

LightGuard: Is it easier to create the environment or hack control.rb and change the composer file? I'll try modifying just the composer file and see what happens, but I'm not 100% sure it'll work that way (but I haven been wrong before).

LightGuard:
@robpblake Take a look at this now, I believe I've covered all that needs to happen (save creating the jenkins jobs and tying them together).

PR #1219, opened 2016-06-26
Developer 3241

Dantheman720:


Dantheman720:
I'll check in later on this and make any necessary changes.

_layouts/product-index.html.slim
```diff
@@ -39,6 +39,11 @@ layout: technology-base
                         - unless i >= (solutions.keys.size - 1)
                           ' ,
               .large-8.columns
+              - if(!product.has_get_started_page)
+                a.button.caps(id="get-started-with-#{product.id}" href="#{site.base_url}/products/#{product.id}/download/") Get Started
+              - elsif(!product.has_get_started_page)
```
paulrobinson: I think you mean: elsif(!product.has_download_page)

Dantheman720:
@paulrobinson I think this will work better. There were some strange styling issues with the way I was doing it before. This should prevent that from happening.

PR #1228, opened 2016-06-27
DTS 4.1 updates

rkratky:
Please merge updates reflecting the current version of DTS (docs links, product version, component names, language fixes).

### DIFF
--- a/products/developertoolset/_common/product.yml
+++ b/products/developertoolset/_common/product.yml
@@ -1,7 +1,7 @@
 name: Red Hat Developer Toolset
 abbreviated_name: Developer Toolset
 product_logo: portal.png
-current_version: "4.1"
+current_version: "4.1.0"
 indented_product: false
 documentation_path: Red_Hat_Developer_Toolset
 product_category: it-infrastructure
--- a/products/developertoolset/_common/product.yml
+++ b/products/developertoolset/_common/product.yml
@@ -1,7 +1,7 @@
 name: Red Hat Developer Toolset
 abbreviated_name: Developer Toolset
 product_logo: portal.png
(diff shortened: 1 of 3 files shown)


LightGuard:
this is ok to test

paulrobinson:
this is ok to test

RubyGemTSL:
retest this please

Dantheman720:
retest this please

LightGuard:
retest this please

LightGuard:
retest this please

rkratky:
@LightGuard, could you please check why is the first check failing?

products/developertoolset/_common/product.yml
```diff
@@ -1,7 +1,7 @@
 name: Red Hat Developer Toolset
 abbreviated_name: Developer Toolset
 product_logo: portal.png
-current_version: 3.1.0.GA
+current_version: 4.1
```
LightGuard: My guess is it's failing because of this line. We're expecting a string and this is a float. You could certainly surround it by quotes, but shouldn't it include a .GA or something like that?

rkratky: Thank you. I suspected as much from looking at the log but wasn't very confident.

No, it shouldn't contain GA -- in fact, no RH product that I know of includes the letters "GA" in its version. The z-number is also used very rarely. I'm pushing a commit with quotes.

LightGuard:
Looks like we're expecting the z number as well.

rkratky:
Z-number added. But this isn't ideal -- the product's official title and version number don't use that, so it can be confusing. We should be able to honour the versioning format used by the product. I'll submit an RFE to change this behaviour.

PR #1231, opened 2016-06-28
https://issues.jboss.org/browse/DEVELOPER-3234

insectengine:
Update promos for the RHD homepage


paulrobinson:
this is ok to test

PR #1233, opened 2016-06-28
fixed more broken tests

RubyGemTSL:


RubyGemTSL:
@robpblake ready now :)

PR #1255, opened 2016-07-07
Developer-3212 - Review all URLs created as part of the content import process into Drupal

Dantheman720:


RubyGemTSL:
retest this please

RubyGemTSL:
@Dantheman720 it's always you that the time zone issue with registration fails :)

It's on my TODO list to fix, sorry.

RubyGemTSL:
retest this please

_ext/jboss_developer.rb
```diff
@@ -411,6 +412,7 @@ def create_page(page, content)
       # Returns a Ruby Hash of the HTTP payload.
       def create_payload(page, content)
         path = create_path page
+        puts "Creating page #{path} from content #{page.output_path}"
```
robpblake: @Dantheman720 We can drop this line if we have the other two log statements I think.

_ext/jboss_developer.rb
```diff
@@ -387,10 +387,11 @@ def exists?(page)
       #
       # Returns the Faraday::Response
       def update_page(page, content)
-        payload = create_payload page, content
         path = create_path page
+        puts "Updating page '#{@base_url}#{path}' with content '#{page.output_path}'"
```
LightGuard: Is this debug info, or are we keeping it in?

LightGuard: If we want to keep it in, I suggest we do some logging here instead of simply `puts`

_ext/jboss_developer.rb
```diff
@@ -399,6 +400,7 @@ def update_page(page, content)
       #
       # Returns the Faraday::Response
       def create_page(page, content)
+        puts "Creating page '#{@base_url}#{create_path(page)}' from content '#{page.output_path}'"
```
LightGuard: Same here

PR #1266, opened 2016-07-12
Fixing DEVELOPER-3253 (no sitemap.xml in export)

LightGuard:
Please see https://issues.jboss.org/browse/DEVELOPER-3253 for full
information.

This saves the sitemap.xml file to the export directory whenever an
export is triggered. Change to the existing test also verifies this
behavior.


robpblake:
@LightGuard The generated sitemap.xml needs to contain the publicly visible endpoint of the site so it needs to be generated from the list of links fetched from the existing sitemap. For example, the sitemap on the production Drupal instance will contain links like this:

```
<url>
  <loc>
     http://developer-drupal.web.prod.ext.phx2.redhat.com/about
  </loc>
  <priority>0.5</priority>
  <lastmod>2016-07-12T16:19:03-07:00</lastmod>
</url>
```

However in the exported sitemap, those links need to be re-written to the public endpoint of the site. E.g for production this would need to look like:

```
<url>
   <loc>
     http://developers.redhat.com/about 
   </loc>
   <priority>0.5</priority>
   <lastmod>2016-07-12T16:19:03-07:00</lastmod>
</url>
```

So we need a way to pass in the final endpoint of the site to the export process and then use that during the generation of the sitemap.xml.

robpblake:
@LightGuard I'd suggest just hard-coded the generated sitemap.xml to developers.redhat.com to keep it simple. What do you think?

LightGuard:
Yep, I don't see any reason for doing anything else. It's only needed in
production.

On Wednesday, July 13, 2016, Rob Blake <notifications@github.com> wrote:

> @LightGuard <https://github.com/LightGuard> I'd suggest just hard-coded
> the generated sitemap.xml to developers.redhat.com to keep it simple.
> What do you think?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/redhat-developer/developers.redhat.com/pull/1266#issuecomment-232375661>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAB1E0gt62H6TddUIV_IMwwg_iZBmsP4ks5qVPhbgaJpZM4JK3HW>
> .
>


-- 
Sent from Gmail Mobile

LightGuard:
Closing until Jenkins comes back

robpblake:
retest this please

LightGuard:
retest this please

robpblake:
@LightGuard Just tested this out and there are a few issues.

1. The URL substitution makes the sitemap.xml invalid because it also replaces namespace declaration at the top of the sitemap e.g:

```
<urlset xmlns="http://developers.redhat.com/schemas/sitemap/0.9" xmlns:xhtml="http://developers.redhat.com/1999/xhtml">
```

2. The sitemap.xml is not copied to the correct location within the export of the site. Currently after running the export I see the following directory structure created by `httrack`:

```
backblue.gif
docker
fade.gif
hts-cache
index.html
sitemap.xml
url-list.txt
```
The sitemap.xml should be in-side the `docker` directory (which is the name of the host the export was run against). 

I would be tempted to create another class that is delegated to by `HttrackExportStrategy`. `HttrackExportStrategy` receives the file containing the list of links that are going to be exported from the Drupal instance. Once the export is complete, it also has the final destination as to where `httrack` has exported to. The new class could be passed the list of links and the export location and generate the sitemap.xml in that directory, substituting `developers.redhat.com` only for the links within the list of links that have been exported.

robpblake:
Tested and confirmed this is working locally.

PR #1271, opened 2016-07-13
Updated the context of development support

dbhole:
Specified what type of development (applications that work with JBoss MW) is supported.


paulrobinson:
I'm going to try manually triggering this build, as the CI server is still down.

PR #1273, opened 2016-07-14
DEVELOPER-3273 - Updated products to no longer list GA variables

Dantheman720:
All products should now show up without GA; however, I need to review Blinkr to make sure that this doesn't cause problems with the documentation.

### DIFF
--- a/_config/blinkr.yaml
+++ b/_config/blinkr.yaml
@@ -35,6 +35,8 @@ skips:
 ignores:
   - code: 202
   - code: 203
+  - code: 403
+  - code: 502
   - snippet: !ruby/regexp /f\.vimeocdn\.com/ # I don't see this error in real browsers
   - url: https://library.oreilly.com/book/0636920025368/continuous-enterprise-development-in-java/toc # works fine in a browser, not an error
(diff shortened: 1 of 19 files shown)


LightGuard:
Closing until Jenkins comes back

LightGuard:
@Dantheman720 Blinkr built, there's 1700-something errors. I think that's a few more than what we were at before, but not sure how much.

LightGuard:
Didn't they also want to possibly drop the final digit?

Dantheman720:
@LightGuard - That was the original request, but the ticket was updated to only drop the GA. I can work on dropping the final digit as well, but we'll need the DoD updated to reflect that.

paulrobinson:
We are just dropping the GA. Dropping the micro version wasn't worth the effort.

LightGuard:
Okay, that's fine then. Do we have sign off for this now?

LightGuard:
There are significantly less issues in blinkr now, I think we are good to merge this, anyone want to give it a second opinion?

Dantheman720:
I think this is ready too. I asked Robert to review and make sure that everything looks okay.

https://issues.jboss.org/browse/DEVELOPER-3273

Dantheman720:
@LightGuard @paulrobinson Robert approved this.

https://issues.jboss.org/browse/DEVELOPER-3273

PR #1275, opened 2016-07-15
DEVELOPER-3288 - Removing the display_submitted text

LightGuard:
All of the created node types should have this set to false, looks like
this one missed it.

I fixed this in production drupal, but this should fix it even if creating a new drupal instance from scratch.


robpblake:
Tested locally and this resolves the issue.

PR #1284, opened 2016-07-19
[DEVELOPER-1998] Support rollback to a previous export

robpblake:
This pull request makes it possible to rollback the site to a previous export. It does this by:

- Ensuring that all successful rsync exports are archived on disk
- Providing a control.rb option to allow the user to specify an export archive to rsync in place of the existing site

This update also performs house-keeping on archives created. Only the most recent 15 export archives are kept by default.


robpblake:
@LightGuard I would appreciate a general review if you have time. The accompanying README.md provides examples of how the script is used.

I don't currently have a way to test this "for real" until we get a successful export in drupal-production or drupal-staging environments.

cheers,

Rob

_docker/lib/rollback/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+Using the naming structure, you can then determine which export you would like to rollback to. If you wish you can also browse into the given export directories to perform
+a manual inspection of the site at that point in time.
+
+### How do I perform the rollback
+
+To use the rollback, you need to execute the script as follows:
+
+```
+ruby rollback.rb <rsync_destination> <name_of_archived_export>
```
LightGuard: Should these be `bundle exec ruby ...`?

_docker/tests/export/test_export/containers.html
```diff
@@ -0,0 +1,2 @@
+<h1>Containers</h1>
+<p>They always work and solve all your problems!</p>
```
LightGuard: I like the sense of humor here

LightGuard:
I don't see anything wrong with this.

PR #1287, opened 2016-07-20
[DEVELOPER-3317] Fix production export

robpblake:
This PR fixes an issue with the export from the production and staging Drupal instances. With this fix hopefully all pages should now be included and re-written correctly.

### DIFF
--- a/_docker/lib/export/drupal_page_url_list_generator.rb
+++ b/_docker/lib/export/drupal_page_url_list_generator.rb
@@ -42,6 +42,7 @@ class DrupalPageUrlListGenerator
       links << link.content
     end
 
+    links.uniq!
     # Ensure that we also grab robots.txt which will not appear in the sitemap.xml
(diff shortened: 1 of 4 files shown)


_docker/lib/export/drupal_page_url_list_generator.rb
```diff
@@ -42,6 +42,7 @@ def parse_sitemap(sitemap_contents)
       links << link.content
     end
 
+    links.uniq!
```
LightGuard: I wouldn't have thought we'd need this, but whatever.

robpblake: We seem to have duplicate links in the sitemap.xml for some reason (particularly relating to quickstarts)

robpblake:
@LightGuard If you get the time to review this today that would be appreciated.

cheers,

Rob

PR #1288, opened 2016-07-20
 Remove Windows from Mac/Linux instructions

paulrobinson:


paulrobinson:
Actually, neither 'windows' or 'cygwin' are mentioned on http://developers.redhat.com/products/cdk/get-started/#tab-macLinux. I think this has been resolved by another PR.

PR #1291, opened 2016-07-21
Updated references for JDG 7.0.0 GA

huffmanca:
JDG 7.0.0 GA'd yesterday, and I have updated the references to point to the latest docs.  Note that the Infinispan Query guide has been removed, while a few additional books have been added to this list.


LightGuard:
this is ok to test

PR #1296, opened 2016-07-22
[DEVELOPER-3324] Workaround default hostname in drupal sitemap.xml

robpblake:
This PR works around the fact that at times, the hostname in the Drupal sitemap.xml gets re-written to be 'default'. It parses the list of links from the sitemap.xml and then ensures that it replaces the 'default' hostname with the one specified in the initialisation of the export process.


robpblake:
@paulrobinson / @LightGuard Could one of you please review and merge if happy?

cheers,

Rob

PR #1306, opened 2016-07-26
[HOLD, PREVIEW ONLY] StackOverflow code migration to AngularJS

adelasofia:


_layouts/base.html.slim
```diff
@@ -367,6 +367,7 @@
       script src="#{site.base_url}/javascripts/a-b-testing.js"
       script src="#{site.base_url}/javascripts/stackoverflow.js"
       script src="#{site.base_url}/javascripts/product-forums.js"
+
     - unless page.extra_javascripts.nil?
```
LightGuard: We need this, don't remove it.

adelasofia: @LightGuard  I didn't remove it, the red back-ground color is coming from Github's error-highlighting feature in the text editor/viewer 😃

LightGuard: Got it. That's what I get for going to be at 3 AM.

LightGuard:
Uh, why was it closed?

PR #1314, opened 2016-07-27
[DEVELOPER-3330] Broken sign-up link on contributor page in Drupal

LightGuard:
See issues.jboss.org/browse/DEVELOPER-3330 for the full info.

We had another relative link, well two actually. I fixed both of them.


robpblake:
retest this please

robpblake:
Tested locally and this seems to work fine.

PR #1320, opened 2016-07-28
[DEVELOPER-3323] Added re-write of form action URLs as part of export

robpblake:
This PR adds in the re-write of form target URLs to the export process if the target URL for the form matches the Drupal host from which it was exported.

### DIFF
--- a/_docker/lib/export/export.rb
+++ b/_docker/lib/export/export.rb
@@ -7,6 +7,7 @@ require_relative 'static_export_rsync'
 require_relative 'export_inspector'
 require_relative 'export_archiver'
 require_relative 'export_archive_pruner'
+require_relative 'form_action_rewrite'
 
 #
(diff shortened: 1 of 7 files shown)


robpblake:
@LightGuard Would you mind taking a look at this one.

I've tested the export locally and it re-writes the pages as expected.

cheers,

Rob

LightGuard:
I discovered an issue with Nokogiri and angular code in `href` attributes. We need to test for that. In your tests add an anchor tag with something like `href="{{blah.test[0]]}}"` and make sure that's what you're still getting after the export. You'll probably need to run the whole string through `CGI::unescape` to make that happen.

robpblake:
@LightGuard Included your recommended CGI::unescape work-around.

cheers,

Rob

robpblake:
retest this please

PR #1321, opened 2016-07-28
[DEVELOPER-3326] Retry logic for cron

LightGuard:


LightGuard:
retest this please

robpblake:
@LightGuard Please see my last comment on https://issues.jboss.org/browse/DEVELOPER-3326

The problem is not to do with the sitemap.xml fetch, its to do with the invocation to the cron URL whilst cron is already running on the Drupal instance. 

That has a default timeout of 10 minutes and will fail if Drupal is already running cron itself (that is my theory). 

My suggestion is that we simply configure Drupal to not run cron itself and rely on the fact that we run cron ourselves at least once per-hour, and have the ability to trigger it remotely if we ever needed to.

LightGuard:
@robpblake shall I apply this for the cron query or drop it and say we're going to fix it by not running cron automatically?

robpblake:
@LightGuard I would drop this one as it does not fix the actual underlying issue which is that cron is running on it's own automatically.

PR #1322, opened 2016-07-29
[DEVELOPER-3360] Switching to angular ng-href

LightGuard:
There's no ticket for this, but it removes some hacks we've had to put in place for the export.

@robpblake, @adelasofia please go over this and verify.


robpblake:
@LightGuard I've run the export locally using this and I believe it solves the issue raised by DEVELOPER-3360.

@adelasofia Can you give this the once over from an Angular perspective to make sure its not going to cause us any problems?

cheers,

Rob

PR #1328, opened 2016-08-02
[DEVELOPER-3366] Navigation is broken while in Admin

LightGuard:
See https://issues.jboss.org/browse/DEVELOPER-3366 for more information.

This moves the page top section in Drupal to the very top of the page,
above our navigation.


robpblake:
Tested locally and this seems to resolve the issue reported within DEVELOPER-3366.

PR #1332, opened 2016-08-02
test fix for prod AC tests

RubyGemTSL:


RubyGemTSL:
@LightGuard this will fix the failing acceptance tests on production build.

PR #1335, opened 2016-08-04
[DEVELOPER-3359] Navigation is broken on search page of site export from Drupal

robpblake:
This PR does a couple of things:

1.  Updates the form target re-writing logic to set the URL relative to the current page rather than absolute. It does this by requiring an anchor element on the same page with the id "home-link", which points to the home page of the site. It then uses the href attribute of this link to set the location of the search page (it makes the assumption that search will continue to exist at /search.html on the site). I've updated the Drupal layout to include this link.

2. Modifies the search.js implementation to rely on Javascript's window.location.pathname attribute when pushing navigation history into the browser. Previously it was hard-coded to '/search/', which was breaking the export of the site from Drupal.

Please fully test the search functions of the site preview before merging this as I have very little Angular experience.


RubyGemTSL:
retest this please

robpblake:
@LightGuard This should hopefully fix the issues that we've been having with the search page navigation in the export from Drupal. Could you give it a the once over please.

@adelasofia Could you take a look at the changes I've made on the Angular side of things. They are quite minimal, but if you could double-check that I haven't done anything silly, that would be appreciated.

cheers,

Rob

robpblake:
retest this please

RubyGemTSL:
retest this please

LightGuard:
retest this please

adelasofia:
@robpblake looks good to me

robpblake:
retest this please

robpblake:
retest this please

LightGuard:
I still can't test this locally. root owns the export directory, not awestruct. We'll see what happens to the prod build when I merge it.

PR #1342, opened 2016-08-05
[DEVELOPER-3370] Registration fails in Drupal and export of site from Drupal

robpblake:
This PR makes a small modification to drupal-namespace.js to support registration in Drupal and the site export from Drupal.

I have tested this manually on a local Drupal instance and a local export from Drupal.

AFAICT this should have zero impact on the non-Drupal version of the site as I have only changed drupal-namespace.js. However please fully test the registration process on the site preview.


robpblake:
This isn't quite right, please don't merge it.

LightGuard:
@robpblake Let @Dantheman720 and I know when this is good to go and we'll get it tested out.

robpblake:
@LightGuard / @Dantheman720 This is now ready to test. You should now be able to complete the registration process from Drupal and from and export of the site.

If you could give it a go I would appreciate it. 

It should have zero impact on the current version of the site as I have only changed drupal-namespace.js.

cheers,

Rob

PR #1343, opened 2016-08-05
[DEVELOPER-3359] Site navigation is broken on search page of export

robpblake:
Another attempt to fix the site navigation from the /search page of the export.

Search functionality on the preview site should be checked thoroughly.

### DIFF
--- a/javascripts/search.js
+++ b/javascripts/search.js
@@ -236,7 +236,7 @@ function searchCtrlFunc($scope, $window, searchService) {
     $scope.query = $scope.params.query; // this is static until the update re-runs
     var params = $scope.cleanParams($scope.params);
     if(isSearch) {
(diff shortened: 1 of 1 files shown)


LightGuard:
@robpblake I cannot export for whatever reason (permission issues again, I disabled SELinux and the export folder is owned by root now). @Dantheman720 maybe you could test this one?

robpblake:
@Dantheman720 Are you able to give this one a test? It will need:

1. control.rb -e drupal-dev --run-the-stack
2. control.rb -e drupal-dev --export

From there, visiting any of the exported pages at http://docker:9000, you should be able to use the search function. From the search results page you should be able to successfully navigate around the site without immediately getting a 404.

Additionally could you please double-check the search function on the preview site as part of this PR to make sure I haven't broken anything there. I have given it the once over and it seems fine, but would be good to have someone double-check.

cheers,

Rob

PR #1352, opened 2016-08-10
Fixes DEVELOPER-3368 (SO forums not working)

LightGuard:
We ultimately need to find the right Adobe DTM code, as what we have
isn't working, or if we're even using DTM any longer.


robpblake:
retest this please

robpblake:
@LightGuard Is this good to merge or are you chasing up whether or not DTM is still be used?

LightGuard:
Still chasing.

LightGuard:
@robpblake This is good. We currently don't have DTM setup for the site, per Marcia Coman. I've removed it from the rhd.settings.yml file for prod.

PR #1353, opened 2016-08-10
Hacking around vimeo limits

LightGuard:
I'm attempting to hack around vimeo limits by using a local cache that
we'll pull from a know location in the build pipeline.


LightGuard:
retest this please

PR #1354, opened 2016-08-11
[DEVELOPER-3389] Fixes Drupal not booting in production

robpblake:
Ensures that we automatically answer 'yes' to any prompt from the drush updb command. This prevents the update process from failing otherwise.


robpblake:
@LightGuard Small update that passes through a "yes" flag to the drush command to update modules. Without this the deployment fails as it asks whether or not to proceed with the update.

PR #1359, opened 2016-08-12
[DEVELOPER-3414] Exclude .snapshot directory from rsync

robpblake:
This PR ensures that we exclude the .snapshot directory from the rsync process. Currently this is causing rsync to staging and production file system locations to fail.

In the absence of being able to integration test the rsync in a local dev environment, I have tested this manually in the staging environment and confirmed that it works as expected.


robpblake:
@LightGuard A very small update to the rsync process to ensure that we exclude .snapshot directories from the process so that we can copy stuff onto directories that are backed up by IT.

PR #1360, opened 2016-08-12
Changed "Windows 8" to say "Windows"

dbhole:
The installer should work on any version of Windows, there is no need to say "Windows 8"


LightGuard:
this is ok to test

PR #1369, opened 2016-08-18
DEVELOPER-2851 Added branded error pages

adelasofia:


paulrobinson:
@adelasofia can you ask James to review?

RubyGemTSL:
@paulrobinson - we probably should have a task in JIRA to test error branded pages now we have them? This is something that could be tested once we start using a web proxy to stub results.

robpblake:
@paulrobinson Just noticed this issue. Additionally it is probably worth understanding how this applies to the Drupal version of the site (in particular the export).

LightGuard:
For Drupal we'll want to edit the System -> Basic Site Settings to add these to the 404 and 403 sections.

LightGuard:
Can't test this because there's no rules in apache currently to display these pages.

PR #1375, opened 2016-08-23
[DEVELOPER-3445] Fixes for Drupal pull requests

robpblake:
Small changes to ensure that Drupal pull requests can work.

1. Needed to install the octokit Gem into the export container as a dependency which running the export with the GitHub status wrapper
2. Ensured that we bind all required environment variables when running the export in the PR environments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# Red Hat Developers Site!
+# Red Hat Developers Site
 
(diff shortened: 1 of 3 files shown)


robpblake:
@LightGuard Could you please have a quick review and merge if you're happy. The extra statuses on the issue are from my earlier testing of the Drupal Pull Request job.

PR #1376, opened 2016-08-23
[DEVELOPER-3356] Fix incorrect bind-mount location for Drupal config 

robpblake:
This PR fixes the bind-mount location for the active configuration directory in Drupal in both staging and production environments.


robpblake:
@LightGuard This PR fixes the incorrect locations of the bind-mount for the active configuration directory in staging and production.

cheers,

Rob

PR #1377, opened 2016-08-23
DEVELOPER-3419 and DEVELOPER-3415

LightGuard:
These could be two separate PRs, but I figured they're both small and dealing with the same thing.


robpblake:
Tested locally and seems to work as expected.

PR #1378, opened 2016-08-23
FIXES DEVELOPER-3395 and DEVELOPER-3397

LightGuard:
3397 was actually done earlier, you can see it live in the production /
staging version of drupal now.

3395 required some small configuration changes.


robpblake:
Tested locally and confirmed that the changes are correct and working as expected.

PR #1379, opened 2016-08-23
Fixes DEVELOPER-3413

LightGuard:
See https://issues.jboss.org/browse/DEVELOPER-3413 for full details.

This should fix it, though there may be something we need to do for
production (maybe rebuild from scratch?).


robpblake:
Tested locally and confirmed there is no image field on the article any more.

PR #1389, opened 2016-08-25
[DO NOT MERGE] ignoring page title check for drupal

RubyGemTSL:


paulrobinson:
@RubyGemTSL are you happy with this run of the acc tests? Does this result tell us that there are no other test failures, other than the page title problems? 

I am sceptical, as I think the /stack-overflow page should have failing tests.

RubyGemTSL:
@paulrobinson sorry I forgot to ping you yesterday. I caught up with Rob, I am happy with this test run. The test failures I have locally are on Login/Registration, SO, Search, and Resources pages.

The failures on this PR are similar: Login/Registration, SO, Search, and Resources pages:

Login/Registration - we know about, keycloak doesn't seem to work as you see Login/Logout/Registration links on the screen shot.

SO page: doesn't load any results.

Search page: doesn't load any results.

Resources page: All tests failed (locally they only failed when filtering results). So i'll need to look into this page in more detail.

paulrobinson:
Ah I see now. I was looking at the wrong test results.

PR #1391, opened 2016-08-25
Fixing 3356 and 3357

drupalbee:


paulrobinson:
Commented on the JIRA

drupalbee:
@paulrobinson & @LightGuard I’ve pushed the changes, and it isn't breaking the website now. But the path auto module isn't working properly and the checkbox isn't showing up. It does after I do another “drupal config:import”. Previously I had added it twice in the script before the file workflow changes were merged but I believe if we do that the website may break since when the path auto module is enabled during the content build stage the page urls change.

LightGuard:
There's an update to drupalconsole which fixes an existing bug with config importing, but you also have to update metatags. Not a big deal. I've done both of those and now I'm stuck with the following message:

```
An error accurred while trying to write the config file: "The import failed due for the following reasons:     
Site UUID in source storage does not match the target storage. 
Entities exist of type <em class="placeholder">Shortcut link</em> and <em class="placeholder"></em> <em class="placeholder">Default</em>. These entities need to be deleted before importing. 
Entities exist of type <em class="placeholder">Workspace</em> and <em class="placeholder">Workspace type</em> <em class="placeholder">Basic workspace</em>. These entities need to be deleted before importing. 
Configuration <em class="placeholder">devel.settings</em> depends on the <em class="placeholder">Devel</em> module that will not be installed after import. 
Configuration <em class="placeholder">pathauto.settings</em> depends on the <em class="placeholder">Pathauto</em> module that will not be installed after import. 
Configuration <em class="placeholder">system.action.pathauto_update_alias_node</em> depends on the <em class="placeholder">Pathauto</em> module that will not be installed after import.
Configuration <em class="placeholder">system.action.pathauto_update_alias_user</em> depends on the <em class="placeholder">Pathauto</em> module that will not be installed after import."  
```

drupalbee:
I have seen this before. This happens when the import process thinks that the website which the config files in the sync folder has is different. I suggest you delete the sync folder, build the website and the import the config file as a whole through the UI. After that export it and bring back the sync folder with those config and things should be back to normal.... Another way is to update the uuid in the sync folder with the one that we have currently built and rebuild again...

PR #1402, opened 2016-08-31
[DEVELOPER-3464] Support HTTPS for site export

robpblake:
This PR adds the ability to define the protocol over which the DCP should be accessed. This means that environments that need to support HTTPS, can be configured to use the HTTPS endpoint of the DCP to avoid browser security warnings.


paulrobinson:
Don't you need to add "protocol: 'https'" to other environments?

robpblake:
@paulrobinson Properties for other environments are not checked-in to GitHub as they contain sensitive credentials to production systems. They are currently stored directly on the production/staging systems were only those with the requisite access can get at them.

PR #1403, opened 2016-08-31
Updated to use friendly YouTube channel URL

paulrobinson:


robpblake:
@paulrobinson This is just an update the contributing.MD file, so I assume that this does not affect the site in any way at all and can be merged?

paulrobinson:
agreed. you should check the URL goes to the same destination as the one I replaced.

PR #1408, opened 2016-09-01
[URGENT] DEVELOPER-3467 Fixes broken Homepage hero in Safari

adelasofia:


paulrobinson:
@robpblake should the Drupal job have failed?

paulrobinson:
We should check that this is fixed in the Drupal export before merging, given that the site is now being served through that process

robpblake:
retest this please

PR #1411, opened 2016-09-02
DEVELOPER-3474 - fixing drupal resources tests

RubyGemTSL:
…sts.


robpblake:
Merging with great pleasure!

PR #1412, opened 2016-09-02
Fixing drush and fields

drupalbee:
While attempting to resolve the issue with the field errors.  I had attempted 
to update Drupal console which ended up causing more problems.  What finally resolved the issue was only updating Drush.  This fixed the issue that fields were not getting created. Going forward we’ll have to only use Drush commands to import config changes instead of Drupal console.

Make changes and export the configuration with drush config-export. It will export new changes to config/sync folder.

Remove the UUID line from the field.storage configurations for the newly created field. If there are more than one field being imported, that UUID line needs to be removed from all the files. 

If the field needs to be updated, instead of being created, the target database (production database) will need to be cloned  and the UUID needs to be kept intact to update the appropriate field.


LightGuard:
@drupalbee looks like the UUID was changed for all files instead of being removed

robpblake:
@drupalbee What are the steps we need to take to apply this to the current production Drupal? Can we just merge your changes and re-start the server as normal?

It would also be good to get the Drush commands into drupal_install_checker.rb and unit tested as well. I also don't know if we need to remove the current config:import commands from there also?

drupalbee:
@LightGuard yes those UUID changes were only removed for the field storage file namely field.storage.node.field_author_name.yml. The changed UUID files don't affect anything. I’ll remove those from the branch.

drupalbee:
@robpblake  Merging into master shouldn't cause any issues, but I dont know how the changes will effect the production environment as I have only tested it with rebuilds of docker. If you can tell me how to test changes on my local environment without doing a complete rebuild, I’ll confirm that for you.  I’ll move the drush commands into drupal_install_checker.rb and add unit tests as well.

LightGuard:
@drupalbee what do you need, a db dump?

drupalbee:
@LightGuard yes, that would be enough. I am under the impression that we don't rebuild the databases any more on production. If so I’m not sure if the config import would throw errors anymore when doing merges into the master.

LightGuard:
Correct, we're no longer rebuilding the DB for production.

LightGuard:
@drupalbee Is this ready to go now?

drupalbee:
@robpblake I did a merge and ran into errors because the branch contains the updated drush version. I’ve fixed the merge issues and tested it. There shouldn't be any more issues.

I wasn’t able to successfully move the drush commands into drupal_install_checker.rb and add  unit testing.  I’m not familiar with ruby and kept running into errors. @LightGuard could you take a look at that please?

LightGuard:
@drupalbee Which commands need to change what do they need to change to?

drupalbee:
@LightGuard 
Here are the changes I need 
1. The following lines need to be removed from run-httpd.sh
drush -y cim   
drush cr all   
drupal config:delete active field.storage.node.field_author_name    
drush -y cim  
drush cr all

2. The following lines need to be run as ruby code inside drupal_install_checker.rb
drush -y cim   
drush cr all  

3. This command needs to be run once only once on the production database for this push
drupal config:delete active field.storage.node.field_author_name    
drush -y cim  
drush cr all

drupalbee:
@LightGuard 
sorry for the confusion, Please see corrected commands. 
1. The following lines need to be removed from run-httpd.sh
drush -y cim
drush cr all
drupal config:delete active field.storage.node.field_author_name
drush -y cim
drush cr all

2. The following lines need to be run as ruby code inside drupal_install_checker.rb
drush -y cim
drush cr all
drupal config:delete active field.storage.node.field_author_name
drush -y cim
drush cr all

drupalbee:
@LightGuard While you're adding those commands maybe try adding this as well?  https://gist.github.com/chadrien/c90927ec2d160ffea9c4

LightGuard:
@drupalbee yeah, tried that, didn't work. I think what I'll want to do is simply run drupal locally

LightGuard:
@drupalbee See https://github.com/drupalbee/developers.redhat.com/pull/4

LightGuard:
retest this please

LightGuard:
retest this please

PR #1420, opened 2016-09-07
Fixes DEVELOPER-3480 (robots.txt missing)

LightGuard:
See https://issues.jboss.org/browse/DEVELOPER-3480 for full information.

Added a few methods to fetch the robots.txt from the Drupal instance and
save that as part of the export. Also fixed tests that changed / were
added.

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/robots.txt
+++ b/_docker/drupal/drupal-filesystem/web/robots.txt
@@ -57,3 +57,7 @@ Disallow: /index.php/user/password/
 Disallow: /index.php/user/register/
 Disallow: /index.php/user/login/
 Disallow: /index.php/user/logout/
+# auth and download-manager URLs do not need to be retrieved
+Disallow: /auth/
+Disallow: /download-manager/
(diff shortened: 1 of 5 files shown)


robpblake:
retest this please

robpblake:
@LightGuard Functionality looks fine. 

We still need to port over the previous robots.txt to Drupal as that  included disallows to `/auth` and `/download-manager`. I'm also not sure if content of the robots.txt produced by Drupal makes sense for the site.

LightGuard:
Sure, we can add in those ones, trying to get a different robots.txt just for the exportis not going to be easy, unless we want to store it somewhere else and fetch it that way.

robpblake:
@LightGuard I've just added a comment to https://issues.jboss.org/browse/DEVELOPER-3481 for an alternative approach that would help us solve the problem with the missing `robots.txt` and also the missing `.htaccess`.

robpblake:
Closing this due to alternative approach implemented in #1429

PR #1423, opened 2016-09-07
DEVELOPER-3177 Made footer stick to bottom of page

adelasofia:


LightGuard:
@adelasofia who needs to sign off?

adelasofia:
@LightGuard not sure since I reported this issue. But I've tested the build and it works like it should 👍

PR #1426, opened 2016-09-08
[DEVELOPER-3461] Workspace bug workaround into start-up scripts

robpblake:
This PR puts the fix for the workspace bug into the Drupal start-up scripts so that if we have to re-build the environment, we don't forget to take that step.

### DIFF
--- a/_docker/drupal/scripts/drupal_install_checker.rb
+++ b/_docker/drupal/scripts/drupal_install_checker.rb
@@ -59,6 +59,19 @@ class DrupalInstallChecker
     end
   end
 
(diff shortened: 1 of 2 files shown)


robpblake:
retest this please

robpblake:
retest this please

robpblake:
@LightGuard This is good to review. I've put the SQL statement used to work-around the workspace install bug into the Drupal start-up scripts. It runs in all environments and for environments where it isn't a problem, it is simply a null-op.

cheers,

Rob

PR #1430, opened 2016-09-12
[DEVELOPER-3509] Ensure /register works on site export

robpblake:
This introduces a few changes to register.html.slim to ensure that the in-lined Javascript works as expected in the site export from Drupal.

### DIFF
--- a/register.html.slim
+++ b/register.html.slim
@@ -5,13 +5,31 @@ head
   script
     include javascripts/vendor/keycloak.js
   javascript:
+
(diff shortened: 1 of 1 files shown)


LightGuard:
retest this please

paulrobinson:
The Drupal build of the site works and tests pass. /register works on the export of the site.  

Got a 500 error when building the Awestruct version, which I assume is temporary. Merging as it's an urgent fix.

PR #1441, opened 2016-09-14
[DEVELOPER-3484] Remove Searchisko Containers

robpblake:
This PR removes the searchisko containers from environments where they are no longer required. This helps us reduce the resource usage on the build server in particular.


robpblake:
@paulrobinson / @LightGuard This is good for a review. I've simply removed the searchisko containers from the environments in which they are no longer used. This means we'll be using significantly less resources on stumpjumper for each PR.

cheers,

Rob

LightGuard:
@robpblake we no longer need it for dev?

PR #1446, opened 2016-09-15
DEVELOPER-3108 Adaptive placeholders' fixes

adelasofia:


_layouts/base.html.slim
```diff
@@ -371,6 +371,7 @@
       / script src="#{site.base_url}/javascripts/searchpage.js"
       script src="#{site.base_url}/javascripts/a-b-testing.js"
       script src="#{site.base_url}/javascripts/product-forums.js"
+      script src="#{site.base_url}/javascripts/adaptive-placeholder.js"
```
LightGuard: Since we're now using Drupal, changing this base layout will not do anything. New javascript files need to be added to the Gulpfile.

javascripts/adaptive-placeholder.js
```diff
@@ -0,0 +1,20 @@
+app.adaptivePlaceholder = {
+
+  changeFilledState: function () {
+    var input = $(this);
+    var value = input.val();
+    
+    if(value != '' && value != undefined) {
```
LightGuard: Mind adding whitespace after the `if` please? We really should set up a standard for this and sass files.

LightGuard: I believe this should be `!==` instead of `!=`

adelasofia: @LightGuard  The JavaScript not equal operator is !=

LightGuard: I understand that, however, it doesn't check the types at all and for that reason can really get you into trouble, for example put these into a console:

```
0 == ''
0 === ''
false != ''
false !== ''
```
The number case could come back and bite you in this particular code section. Yes, it's a corner case, but certainly something we should be considering. Food for thought: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness and http://dorey.github.io/JavaScript-Equality-Table/

If you feel like we're good using `!=` here, okay, that's your call.

KamiQuasi: Would definitely recommend avoiding type coercion if at all possible. Still getting into the swing of the workflow, but additionally some linters would balk at the use of single quotes as well.

adelasofia: @LightGuard @KamiQuasi got it, thank you! I've pushed the changes

PR #1460, opened 2016-09-22
Increased button font-size to 16px, added line-height 1.2em

KamiQuasi:
Resolves DEVELOPER-3473 to sync up with other RH properties

### DIFF
--- a/stylesheets/app.scss
+++ b/stylesheets/app.scss
@@ -159,10 +159,11 @@ button, .button {
   text-transform: uppercase;
   text-decoration: none;
   border:0;
-  font-size: 15px;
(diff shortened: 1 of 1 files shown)


cdrage:
Hi @KamiQuasi I came across this while trying to find as to why the font (specifically `p` paragraphs) are so small on developers.redhat.com

Has this been reverted somewhere perhaps? It's *wayyyyy* too small on the site currently for blog posts.

KamiQuasi:
@cdrage this change was strictly for the font size of buttons, and doesn't really impact the text on the rest of the site. I think that the Blog pages would be a separate issue to address.

PR #1469, opened 2016-09-26
added social test to nightly run as is failing when 1 or more tests r…

RubyGemTSL:
…un at the same time. Causing failures on CI.


RubyGemTSL:
retest this please

PR #1485, opened 2016-09-30
DEVELOPER-3468 (Canonical URLs)

LightGuard:


LightGuard:
retest this please

LightGuard:
retest this please

LightGuard:
retest this please

robpblake:
@LightGuard This approach won't work in consideration of the export process. After the export has run, the canonical tag will point to the Drupal URL rather than developers.redhat.com. I'd suggest if possible we simply hard-code the beginning of the canonical URL to be developers.redhat.com if possible, otherwise we're going to need to alter the export process to re-write the canonical URL value.

LightGuard:
Yeah, I think I realized that after I submitted this PR, then did some digging into what this link tag actually does. IMO I still don't think we should do it. Check the JIRA for my reasoning.

paulrobinson:
I'm closing this for now. We can re-open when we've consulted an SEO expert and we know what to do.

PR #1487, opened 2016-10-01
[HOLD] Updated SO query to use the same DCP as the rest of the build.

paulrobinson:
Trying this again with a clean PR. Maybe there was some stale state in the old one.


paulrobinson:
@RubyGemTSL any ideas why these tests are failing: http://jenkins.mw.lab.eng.bos.redhat.com/hudson/job/developers.redhat.com-acceptance-test-docker/2420/artifact/_cucumber/reports/rhd_test_report.html.

They seem to be timing out. Maybe the SO query is taking too long in the staging environment, vs the production environment? That's the only change in this PR.

RubyGemTSL:
@paulrobinson strange that it is not taking a screenshot on fail. Will investigate.

RubyGemTSL:
@paulrobinson - there is one failing resources test that I can reproduce. It is failing on a test that selects 'show results from "Last Quarter"'. It seems the results that are displayed are from slightly over the past quarter. I've raised a PR to ignore this test until we can intercept results from DCP and stub results as it is always failing. The SO pages are working fine for me so maybe there was some issues when the tests were ran. I'll do a retest.

RubyGemTSL:
retest this please

RubyGemTSL:
Once #1493 is merged this will pass.

LightGuard:
retest this please

RubyGemTSL:
retest this please

RubyGemTSL:
retest this please

LightGuard:
retest this please

RubyGemTSL:
retest this please

RubyGemTSL:
retest this please

LightGuard:
retest this please

LightGuard:
Need to see what's going on with this one

RubyGemTSL:
@LightGuard for some reason with this PR I have to log in to stumpjumper and manually kill the pr image in order to retest it. The tests are bombing out for drupal as the previous image is still hanging around.

paulrobinson:
That's odd. Maybe we leave this PR around for @robpblake to investigate on his return? It may be symptomatic of something bigger and might happen again.

robpblake:
retest this please

robpblake:
@paulrobinson / @RubyGemTSL I can't see anything Docker specific causing a problem here. From what I can see the StackOverflow tests time out on original acceptance test job, which means that they are still running when the Drupal Acceptance test job kicks off.

It looks like to me that using the staging DCP with the StackOverflow tests just doesn't work. @RubyGemTSL Please correct me if I'm wrong?

RubyGemTSL:
@robpblake - yep you're right. This is the only PR that is affected by this. As we discussed, shall we just run the smoke tests on awestruct builds? In the unlikely event that we have to switch to the awestuct build we can run the full suite of tests before the switch?

paulrobinson:
I think we'll be turning off the Awestruct builds soon anyway. I'm ok with us switching to just smoke-tests on Awestruct.

paulrobinson:
retest this please

paulrobinson:
retest this please

paulrobinson:
retest this please

paulrobinson:
@robpblake so we finally got this working. Can you review?

You can see it actually working on the site here: http://developers-pr.stage.redhat.com/pr/1487/export/stack-overflow/?q=

Also the tests passed, apart from an unrelated and expected failure.

RubyGemTSL:
retest this please

PR #1495, opened 2016-10-04
[DEVELOPER-3570] Fix Wrapper Class Flex Basis

KamiQuasi:
Flex short-hand caused IE to set the flex-basis property of the .wrapper class to 0%, which pulled the footer up into the content.

https://issues.jboss.org/browse/DEVELOPER-3570

### DIFF
--- a/stylesheets/app.scss
+++ b/stylesheets/app.scss
@@ -468,6 +468,7 @@ body.fullbleed {
   min-height: 400px;
   padding-bottom: emCalc(100px);
   flex: 1;
+  flex-basis: auto;
 }
 
 


LightGuard:
Is this good to go?

KamiQuasi:
Yeah, should resolve the weird IE 11 layout issues. Tested it locally on a VM, but if someone else wants to verify feel free.

LightGuard:
@adelasofia or @Dantheman720 do you have an IE 11 vm you can test with?

RubyGemTSL:
@KamiQuasi @LightGuard - tested on IE 11, and 10. Looks good.

PR #1510, opened 2016-10-11
Force Fuse docs to update to v6.3

paulrobinson:


LightGuard:
Looks like that did it.

LightGuard:
@paulrobinson a test timed out on the normal acceptance tests

paulrobinson:
I guess we need to go with this hack for now? We need the Fuse 6.3 docs live today.

LightGuard:
Yep. Not really a hack, well, I guess it's a workaround.

PR #1539, opened 2016-10-18
[DEVELOPER-3613] Do not delete CSS from receiving side in rsync

robpblake:
This PR updates the rsync command used to copy the static HTML export to the Apache server. It prevents the receiving side from deleting any .css files even if they do not exist on the sending side.

This should mean that all .css files continue to exist on the receiving side forever, which will mean that MouseFlow functionality should work as expected.

I have tested this by running the rsync command locally, but I do not believe we actually have a method of testing the rsync command between two hosts until we push it into production as each pull-request build rsyncs to a different location.


robpblake:
retest this please

LightGuard:
retest this please

LightGuard:
Alright, we'll give it a go.

PR #1540, opened 2016-10-18
Adding Drupal Config Changes to Fix [site:url] linking

KamiQuasi:
Jason made some changes to fix the issue in production that we're having with the [site:url] being truncated on save of content.


LightGuard:
retest this please

LightGuard:
Manual testing, test failure is a false positive

PR #1545, opened 2016-10-19
Fixing false positive test results. Having the report generated using…

RubyGemTSL:
Fixing false positive test results. Having the report generated using the new rake task - after the rerun has occurred produces a status code of 0, therefore jenkins sees that as pass. So it passing the build, even though there may be failures.

### DIFF
--- a/_cucumber/cucumber.rake
+++ b/_cucumber/cucumber.rake
@@ -2,20 +2,11 @@ require 'rubygems'
 require 'parallel'
 require 'fileutils'
 require_relative '../_lib/github'
-require 'report_builder'
 
-task :cleanup do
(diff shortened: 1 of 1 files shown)


RubyGemTSL:
@LightGuard @robpblake - can we get this merged asap. Tests are passing when they should be failing when the rerunner is triggered.  Having the report generated using the new rake task - after the rerun has occurred produces a status code of 0 as the report is successfully generated, therefore jenkins sees that as pass! Therefore passing the build, even though there may be failures!

RubyGemTSL:
retest this please

robpblake:
From what I can see this looks OK and the tests are green. Merging as requested by @RubyGemTSL .

PR #1546, opened 2016-10-20
[DEVELOPER-3621] Ensure that test reports are generated 

robpblake:
It would appear that we are no longer having the test reports being generated as a result of #1545. As well as being inconvenient that we no longer have the reports, it also causes the build to fail each time as the archive of artefacts build step fails. This attempts to resolve that issue.

Before merging this, please click through to the actual acceptance test job and make sure you can browse to the report!

### DIFF
--- a/_cucumber/cucumber.rake
+++ b/_cucumber/cucumber.rake
@@ -2,6 +2,7 @@ require 'rubygems'
 require 'parallel'
 require 'fileutils'
 require_relative '../_lib/github'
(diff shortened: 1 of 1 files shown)


robpblake:
@paulrobinson / @LightGuard Could one of you please review and merge this fix. We are currently not getting any test reports generated which is causing the acceptance test job to fail when it tries to archive the artefacts.

cheers,

Rob

PR #1551, opened 2016-10-24
Updated docs for adding books

paulrobinson:
We no longer want to pull from the Google Books API as it is too unreliable. The docs now mandate that all information is provided.


paulrobinson:
Merging myself (naughty) as it's just a change to the README.md and I need to provide a link to the updated doc.

PR #1553, opened 2016-10-24
Removing Workbench Moderation

LightGuard:
I made some additional changes to what Rob has done.


LightGuard:
This actually worked, but seems like there's lots of extra results being returned.

PR #1555, opened 2016-10-25
[DEVELOPER-3605] Export process should fail if there are pages missing

robpblake:
This PR updates the export process to fail if there are missing pages. Previously the behaviour was to simply emit a warning if there were pages missing from the export, however these were going un-noticed and the site was being published with pages missing.

I have also included a mechanism to override this default failure behaviour if you absolutely need to. By setting the `drupal.export.fail_on_missing` environment variable to `false`, you can disable the failure behaviour. This allows you to push an export in times when bad things are happening and you just need to get something up. 

I've added the `drupal.export.fail_on_missing` variable to the `export` container definitions in docker-compose.yml for each environment. The variable is currently set to nothing meaning that the failure behaviour is enabled by default. Should we ever need to disable the failure behaviour, then we can simply set the value of the `drupal.export.fail_on_missing` variable in the Jenkins job configuration for the export.

### DIFF
--- a/_docker/lib/export/export_inspector.rb
+++ b/_docker/lib/export/export_inspector.rb
@@ -2,6 +2,11 @@ require 'uri/http'
 
 #
 # The purpose of this class is to inspect an export directory to ensure that all pages that should be there, are in fact there.
+# If a page is missing from the export, then an exception is raised causing the export process to fail.
+#
+# The raising of an exception can be overridden if the 'drupal.export.fail_on_missing' system property is set to 'false'. This
+# should only be used in situations where you want the export to continue even if there is content missing e.g. you're having a very
+# bad time of things with Drupal.
 #
 # @author rblake@redhat.com
 #
@@ -18,6 +23,21 @@ class ExportInspector
     else
       "#{page_path}/index.html"
     end
+  end
+
+  #
+  # Determines if the process should fail if content is missing from the export
+  #
+  def should_fail_on_missing_content
+    fail_on_missing = ENV['drupal.export.fail_on_missing']
+    fail = true
+
+    if !fail_on_missing.nil? && !fail_on_missing.empty?
+      if fail_on_missing.to_s == 'false'
+        fail = false
+      end
+    end
+    fail
   end
 
   #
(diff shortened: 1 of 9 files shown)


robpblake:
@LightGuard This updates the export process to fail if there are pages missing. Appreciate if you could give the small code changes the once over and also test the export locally to make sure it works for you as well.

cheers,

Rob

_docker/lib/export/export_inspector.rb
```diff
@@ -21,6 +26,21 @@ def determine_expected_filesystem_path(url)
   end
 
   #
+  # Determines if the process should fail if content is missing from the export
+  #
+  def should_fail_on_missing_content
+    fail_on_missing = ENV['drupal.export.fail_on_missing']
+    fail = true
+
+    if !fail_on_missing.nil? && !fail_on_missing.empty?
```
LightGuard: Wouldn't

```
unless fail_on_missing.nil && fail_on_missing.empty?
```
read a bit easier?

_docker/lib/export/export_inspector.rb
```diff
@@ -21,6 +26,21 @@ def determine_expected_filesystem_path(url)
...
+      if fail_on_missing.to_s == 'false'
```
LightGuard: In ruby you'd typically use a match for this:
```
fail = false if fail_on_missing =~ /false/
```

_docker/lib/export/export_inspector.rb
robpblake: Yep.

_docker/lib/export/export_inspector.rb
robpblake: Thanks.

robpblake:
retest this please

robpblake:
This is now failing because there are pages missing from the export. So at least we know the code works I guess...

robpblake:
@LightGuard I've made the changes you proposed. This is now failing as there are pages missing from the export, so I need to address that first.

robpblake:
retest this please

robpblake:
@LightGuard I've updated the DB dump on stumpjumper to be the latest production backup and this now seems to be working fine. Ready for review and merge when you're ready.

cheers,

Rob

PR #1558, opened 2016-10-27
DEVELOPER-3429 - Event dates not displaying correctly

KamiQuasi:
https://issues.jboss.org/browse/DEVELOPER-3429

This fix pre-supposes that all dates coming into the event moment filter are UTC times and that the day, month, and year are the only relevant data points coming back from the service. If dates are encoded to a specific timezone this may skew the displayed dates.

### DIFF
--- a/javascripts/events.js
+++ b/javascripts/events.js
@@ -140,6 +140,6 @@ dcp.controller('eventsController', function($scope, searchService, helpers) {
 
 dcp.filter('moment',function() {
   return function(dateString, format){
-    return moment(dateString).format(format);
+    return moment.utc(dateString).format(format);
   };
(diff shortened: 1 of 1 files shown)


KamiQuasi:
retest this please

KamiQuasi:
@LightGuard @paulrobinson @adelasofia @Dantheman720 anyone know of any reason why taking the UTC relative date for the data from the events service would not be acceptable? Do we ever have events where we show the time of day?

paulrobinson:
We currently only ask for the start and end date of the event. We don't specify the time in those fields. I think if we did specify time (in the future), we'd use a different field.

PR #1556, opened 2016-10-26
Reverted back to having Awestruct push the Homepage

paulrobinson:
I spoke to Adela, and we agreed that the homepage is too complex to edit in HTML. It would be better to leave this page in Awestruct until either the Awestruct build is turned off, or the homepage is redesigned.

PR #1561, opened 2016-10-28
[HOLD for Monday Oct. 31] DEVELOPER-3626 New .NET Core RHD Hero

adelasofia:


adelasofia:
@LightGuard Any idea why in the site preview I'm getting a combination of the current homepage hero and the new one I'm trying to push? The preview has the right CSS but the HTML was not updated

robpblake:
@adelasofia I suspect that you'll need the change in #1556. There was an update a while back to stop Awestruct pushing in the homepage into Drupal so I suspect your changes to the HTML are being ignored. @LightGuard can probably confirm this better than I, but that is my suspicion.

It may be easier to for to apply the changes in #1556 as part of your PR.

cheers,

Rob

adelasofia:
@robpblake thank you!

adelasofia:
Approved by James -- This should be merged on Monday Oct. 31st

robpblake:
@adelasofia I see a broken image on the following page: http://stumpjumper.lab4.eng.bos.redhat.com:33820/promotions/dot-net-core

robpblake:
@adelasofia Ignore my previous comment. It is because we're not pulling the images down from production into the pull request environment.

PR #1562, opened 2016-10-28
DEVELOPER-3610 Added MicroServices to main menu

adelasofia:


paulrobinson:
Did you see that you also need to update the welcome page? Seems missing from here.

LightGuard:
@adelasofia This needs to be added to html.html.twig

PR #1590, opened 2016-11-09
Removed version number text from WildFly link

paulrobinson:
This link is unlikely to get updated as new versions are released. It also adds little value, so I've simplified the text.

### DIFF
--- a/products/eap/download.adoc
+++ b/products/eap/download.adoc
@@ -3,6 +3,6 @@
 
 == Upstream
 
-http://www.wildfly.org[WildFly 8.1.0, role="download-link"] +
+http://www.wildfly.org[WildFly, role="download-link"] +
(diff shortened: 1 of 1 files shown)


paulrobinson:
@RubyGemTSL can you review this content change? You just need to look at http://developers-pr.stage.redhat.com/pr/1590/export/products/eap/download/ and check that the "Prefer to use the Upstream Project Version?" section looks right. It should simply have the version number removed.

paulrobinson:
I'm not getting a response from the person who requested this. I think this is the right fix though, so would rather merge it and undo if we are ever told it's wrong.

RubyGemTSL:
@paulrobinson - looks fine to me.

robpblake:
retest this please

robpblake:
@paulrobinson Is this good to go?

paulrobinson:
@robpblake yes. See my last comment.

PR #1591, opened 2016-11-09
Added utility scripts and docs

paulrobinson:
These scripts are really useful. I didn't want them to just live on my laptop, so I thought they should be documented and put somewhere for others to use. 

Is this the right place? If not, where would be better?

### DIFF
--- a/_bin/Readme.md
+++ b/_bin/Readme.md
@@ -30,11 +30,13 @@ To find a list of pages with duplicate titles or descriptions:
       
 E.g. To find duplicate titles:
 
-      python3 ./duplicates.py "https://dcp2.jboss.org"  "title" $dcp_super_user $dcp_super_user_password
+      python3 ./duplicates.py "https://dcp2.jboss.org"  "title" $dcp_super_provider $dcp_super_provider_password
       
 E.g. To find duplicate descriptions:
 
-      python3 ./duplicates.py "https://dcp2.jboss.org"  "description" $dcp_super_user $dcp_super_user_password
+      python3 ./duplicates.py "https://dcp2.jboss.org"  "description" $dcp_super_provider $dcp_super_provider_password
+      
+_Note:_ The DCP SuperProvider password can be found here: https://mojo.redhat.com/docs/DOC-85791
       
       
 ## Find Web Pages with Missing Meta-Data needed by /search
--- a/_bin/Readme.md
+++ b/_bin/Readme.md
@@ -1,0 +1,51 @@
+# Utility Scripts
+This directory contains several utility scripts.
+
+## Remove Stale Content From the DCP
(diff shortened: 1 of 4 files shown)


LightGuard:
Looks like we have another potentially always broken KC test

paulrobinson:
I think we can merge this without the KC tests passing.

@LightGuard happy to have these scripts in this location?

robpblake:
retest this please

robpblake:
@paulrobinson I don't see there being any issue with merging this as it won't affect the actual site. 

My major comment is about commonality throughout the project. It would be nice to standardise all scripts/utilities around one language and I think the preferable one there is Ruby given as that is what the project is already using and the one that the project has the most available resource for.

My other comment would be that I feel that there should be unit tests with this (particularly if these scripts can be used against production systems). That again adds to the common language argument as we don't need to integrate another test suite and could include these utility scripts in the test suites for the other growing body of scripts associated with the project.

paulrobinson:
@robpblake Agreed on both counts. I inherited these scripts from Rysiek, so there wasn't an opportunity to influence these decisions.

paulrobinson:
I'm not sure we have the bandwidth to re-write them in Ruby. However, i am also nervous about us running untested scripts on production servers.

I can ask Rysiek to write unit tests before we merge this.

paulrobinson:
@unibrew I'm adding the scripts you gave me to the project. However, we are a bit nervous about running un-tested scripts on the production DCP. Especially given that one of them deletes content. 

Are you able to write some tests for these scripts?

unibrew:
Sorry for not using Ruby, it was not expected that those scripts will become part of your build. I thought that someone will only run them occasionally. And since I'm not a Ruby fan python came to the rescue ;).

It should be possible to write some tests but probably not during this sprint and not sure about the next one atm. I need to get some things done before December freeze of releases.

Also since those scripts are under 70 lines long each and half of that code is parsing parameters there is not much to unit test. Much more valuable would be integration testing with some DCP instance. As far as I remember you create some DCP instance for testing purposes, am I right ? Something where I could push some content, query and delete.

paulrobinson:
@unibrew sure, I get it. The scripts were only intended to be ran once to get an initial report. 

I'm not really sure how to test this. The DCP Docker instance we have is missing any data that is added via a River. This is problematic as it means it is missing the rht_website content. 

Maybe we create an account on the production DCP that has read-only access? We could force the scripts to use that account.

As for the deleteStaleDocuments.py, I suggest that is left out. It needs write access to the production DCP and is hard to test. Not a good combination! Also, we are overhauling the process by which we populate the DCP, so stale data will be avoided. Thus the usefulness of this script is short-lived.

PR #1596, opened 2016-11-11
DEVELOPER-3643 - Implement User logic

KamiQuasi:
https://issues.jboss.org/browse/DEVELOPER-3643

Creates async user login from keycloak information includes keycloak ID, social accounts linked, and days since registration


KamiQuasi:
@paulrobinson would you give this a look and merge if possible?

PR #1599, opened 2016-11-14
DEVELOPER-3469 - Google Tag Manager script HTTPS always

KamiQuasi:
https://issues.jboss.org/browse/DEVELOPER-3469

Updates the GTM include to be HTTPS-only, to prevent protocol incompatibilities between secure and non-secure access.


paulrobinson:
retest this please

PR #1602, opened 2016-11-14
DEVELOPER-3654 - White cards missing from bottom of Topics pages

KamiQuasi:
https://issues.jboss.org/browse/DEVELOPER-3654

Filtered DCP results to only include valid items onto the page. This prevents the script error from occurring. Validate by visiting /containers and seeing more than four white cards.


RubyGemTSL:
retest this please

KamiQuasi:
@Dantheman720 or @adelasofia will one of you take a look at this by navigating to http://developers-pr.stage.redhat.com/pr/1602/export/containers/ and just make sure there isn't a scripting error, but also that there are enough white cards? Thanks!

adelasofia:
@KamiQuasi It looks good to me 👍

paulrobinson:
Many of those links 404. Might be due to bad data in staging. not sure yet. Also, when you run the same DCP query on production, forum thread results are coming back: https://dcp2.jboss.org/v2/rest/search/developer_materials?tags_or_logic=true&newFirst=true&s...

I don't think we want them in there. I suggest explicitly specifying the content types we want. I think the list is the same as on /resources. You are likely to get web page results too with the current query

KamiQuasi:
Meeting with @RubyGemTSL tomorrow and should have tests written before the end of the day to verify this and get this accepted.

Dantheman720:
@KamiQuasi - Is this ready to go? I can merge first thing tomorrow if so.

KamiQuasi:
@Dantheman720 yeah should be ready to go. I guess take a look and make sure you don't see anything glaringly wrong

PR #1606, opened 2016-11-16
DEVELOPER-3661 Migrated JBD /docker to RHD

adelasofia:


paulrobinson:
The 'More Info' buttons should un-hide some elements. You can see this in action on the original page. I think you may have forgotten to add a JS file.

paulrobinson:
This looks right now. Ready to merge when the tests are passing.

PR #1614, opened 2016-11-21
DEVELOPER-3660 Set content types for product learn pages

adelasofia:


LightGuard:
Are we expecting to see Forum posts in the Learn pages? The label for a forum post is also not correct English, looks like we're dumping the type given from the DCP which is not what people would want to see. Do we also want to add forum posts to the select box of things to limit?

adelasofia:
@LightGuard This PR set the content types that should be returned for the Learn pages, so you should no longer see any Forum links. 

The content types that should be returned, if available, are the same 11 as the Resources page: 

- jbossdeveloper_quickstart
- jbossdeveloper_demo
-  jbossdeveloper_bom
- jbossdeveloper_archetype
- jbossdeveloper_example
- jbossdeveloper_vimeo
- jbossdeveloper_youtube
- jbossdeveloper_book
- rht_knowledgebase_article
- rht_knowledgebase_solution
- jbossorg_blog

LightGuard:
Yep, not seeing that at http://developer-drupal.web.prod.ext.phx2.redhat.com/products/eap/learn

adelasofia:
retest this please

adelasofia:
retest this please

PR #1625, opened 2016-11-29
DEVELOPER-3681 - Investigate & fix failing Resource/Topics tests

RubyGemTSL:
…amically loading elements.

### DIFF
--- a/_cucumber/features/step_definitions/resources_steps.rb
+++ b/_cucumber/features/step_definitions/resources_steps.rb
@@ -25,9 +25,11 @@ When(/^I click to filter results by "([^"]*)"$/) do |filter_type|
   end
 end
 
-Then(/^the results should be filtered by (.*)$/) do |filter|
+Then(/^the results should be filtered by (.*)$/) do |filter_type|
   on ResourcesPage do |page|
-    page.results_contain_img_for(filter).should == true
+    wait_for(30, "Images for #{filter_type} were not displayed after 30 seconds") {
+      page.results_contain_img_for(filter_type).size == 10
+    }
   end
 end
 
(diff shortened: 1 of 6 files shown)


RubyGemTSL:
@paulrobinson - there is 14 white cards displayed on mobile devices. I assume there should be 16 as per desktop view? cc @KamiQuasi

KamiQuasi:
@RubyGemTSL kind of a weird case, but the white card test was mostly just to make sure we had the minimum for the desktop of 16. The visibility of fewer on the mobile view is a function of the style sheet not the actual number of records returned. If we want to bump up the visible number to 16 we just have to increase the height of the container of the white cards.

RubyGemTSL:
Thanks @KamiQuasi - we can either tag with @desktop to only run on desktop or fix for mobile? Do you have an opinion either way on this @paulrobinson ?

paulrobinson:
I think it makes sense to show the same amount of card on mobile vs desktop. Any reason not to?

RubyGemTSL:
OK I raised a bug https://issues.jboss.org/browse/DEVELOPER-3682

RubyGemTSL:
pushed a @desktop only tag for the failing white card tests for mobile. Will be fixed as part of issue I raised above.

RubyGemTSL:
@KamiQuasi any idea why there is only 12 white cards on the .NET topic page? (This is Desktop view too)

KamiQuasi:
@RubyGemTSL it appears that the query being used has some unwanted %20 (space character) in some of the terms like "%20.net" let me see if I can adjust them and that should help.

RubyGemTSL:
retest this please

KamiQuasi:
@RubyGemTSL and @paulrobinson the "fix" was to log in to Drupal and remove the extra spaces from the .Net Topic's "Data Tags" field. The values there are apparently parsed exactly and additional space around the comma-delimited values causes the front-end to query using the space characters, which reduces the number of results.

RubyGemTSL:
@KamiQuasi any idea how/when the changes will make it to PR environment as this is still failing

paulrobinson:
The buzz tags on the .NET topic in production were not updated. I just updated them now. I suggest retrying this in about 3hours.

RubyGemTSL:
retest this please

RubyGemTSL:
retest this please

paulrobinson:
@RubyGemTSL is this good to go now?

LightGuard:
Looks like we're good. @RubyGemTSL ?

RubyGemTSL:
@LightGuard also good - thanks!

PR #1628, opened 2016-11-29
DEVELOPER-3670

weimeilin79:
Update JBDS version


LightGuard:
@weimeilin79 There's a conflict here with `products/fuse/get-started.adoc` please fix before we test.

PR #1635, opened 2016-12-01
[HOLD] [DEVELOPER-2970] Force users to use HTTPS

paulrobinson:
Updating all hardcoded http links to use https. This should be covered by redirects anyway, but it's always better to avoid them where possible. Also, this PR allows us to make sure that export and tests work correctly, in advance of the change.


_config/site.yml
```diff
@@ -211,17 +211,10 @@ profiles:
       <<: *deploy
       path: /stg_htdocs/it-rhd-stg-main
 
-  # Octopus integration configuration
```
KamiQuasi: Just checking since this isn't strictly a protocol change, but is this removal okay?

paulrobinson: Yeah, i removed it as it's redundant and would have required code changes as part of this refactor. It's in a separate commit.

paulrobinson:
Replaced with another PR.

PR #1642, opened 2016-12-05
DEVELOPER-3687 Fixes HTML binding in SO page

adelasofia:


paulrobinson:
This looks to be working: http://developers-pr.stage.redhat.com/pr/1642/export/stack-overflow/#!q=webserver

@KamiQuasi  can you do the code review?

javascripts/search.js
```diff
@@ -201,12 +201,12 @@ search.filter('description', function($sce) {
...
 
-search.filter('question', function($sce) {
+search.filter('question', function() {
   return function(result) {
     if (result.highlight && result.highlight._source.sys_content_plaintext) {
-      return $sce.trustAsHtml(result.highlight._source.sys_content_plaintext[0]);
+      return String(result.highlight._source.sys_content_plaintext[0]).replace(/&lt;/g,"<").replace(/&gt;/g,">");
     }
-    return $sce.trustAsHtml(result._source.sys_content_plaintext);
+    return String(result._source.sys_content_plaintext).replace(/&lt;/g,"<").replace(/&gt;/g,">");
```
KamiQuasi: The best solution appears to be something similar, but containing the replacement for both open and close tags:
`return text ? String(text).replace(/<[^>]+>/gm, '') : '';`

KamiQuasi: OH! Scratch what I said. I was thinking we were trying to remove the tags. This looks like a sticky situation. It almost looks like if we stripped the tags using the code I just posted on the `result._source.sys_content` then it would be valid to use the trustAsHtml function as is. So it would look like 
`return $sce.trustAsHtml(result._source.sys_content.replace(/<[^>]+>/gm, ''));`

adelasofia:
retest this please

adelasofia:
@KamiQuasi thanks! I've made the change you recommended. I believe the failing tests are unrelated to this PR, see Ian's comment in [#1646](https://github.com/redhat-developer/developers.redhat.com/pull/1646) which has the same failing tests

adelasofia:
@KamiQuasi also to make review easier, this was the page that was breaking with the second result https://developers-pr.stage.redhat.com/pr/1642/export/stack-overflow/#!q=webserver

PR #1643, opened 2016-12-05
Adding in apidocs for javadoc

LightGuard:
Adding a new `apidocs` directory and setting it to protect in the rsync.

Hoping this doesn't clobber things already existing.


LightGuard:
@KamiQuasi would you merge this please when it's good? As long as everything is good then it's ready. There isn't anything to test on this outside of the unit tests. It's protecting a folder that actually doesn't exist inside the PR environment.

KamiQuasi:
@LightGuard will do.

PR #1645, opened 2016-12-06
Forcing rebuild of homepage

paulrobinson:
This should remove the mixed-content warnings on the homepage.


paulrobinson:
To review. Visit: https://developers-pr.stage.redhat.com/pr/1645/export/ and check there are no mixed-content warnings in the console log.

I tried it and it looks good.

RubyGemTSL:
There is 1 false failure on this PR (as well as others), the 6th result returned from DCP (when searching for containers) does not contain 'container'.

PR #1660, opened 2016-12-12
Developer 3715 - Cucumber report is not updated to reflect rerun scenarios.

RubyGemTSL:


paulrobinson:
How can I check this works?

RubyGemTSL:
@paulrobinson it's pretty hard to check that it works on a PR. I've verified that it works locally.

To verify locally you would need to run the tests with a deliberate failure, and then when the test run is complete and the rerunner is about to run fix the deliberate failure so it passes on the rerun.

PR #1679, opened 2016-12-20
Fixing DEVELOPER-3737 Connectors displaying twice

LightGuard:
Please see https://issues.jboss.org/browse/DEVELOPER-3737 for full
information and testing.

I decided to comment things out in case we want to bring back featured connectors.


LightGuard:
@Dantheman720 @adelasofia Tiffany said this was good to go

PR #1680, opened 2016-12-21
DEVELOPER-3728 - Adobe DTM User Auth Events Double-counting

KamiQuasi:
https://issues.jboss.org/browse/DEVELOPER-3728

Moves the user auth event outside of the authentication to push whether the user is authenticated or not. May change at some point to a different event name based on Search Discovery's input after the New Year.

### DIFF
--- a/javascripts/sso.js
+++ b/javascripts/sso.js
@@ -33,8 +33,6 @@ app.sso = function () {
                     }
                 }
 
-                updateAnalytics(usr);
-
             }).error(clearTokens);
         } else {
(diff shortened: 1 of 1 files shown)


KamiQuasi:
retest this please

KamiQuasi:
retest this please

LightGuard:
@KamiQuasi is this ready to go?

KamiQuasi:
@LightGuard yep looks like the event is firing now regardless of whether the user is logged in or not, which is what Search Discovery asked for.

PR #1684, opened 2016-12-22
fix for failing test on PR's

RubyGemTSL:


RubyGemTSL:
retest this please

RubyGemTSL:
@adelasofia @KamiQuasi @LightGuard @Dantheman720 

PR's are failing because of the tests for the JBoss redirect to RHD.
The element that we're checking 'referral-alert' seems to now exist twice within the page source, but only on /products/#{product}/overview pages - hence why the tests are failing. Anyone any idea what could have cause this? Does it require fixing or do I need to update the test?

KamiQuasi:
@RubyGemTSL @adelasofia @LightGuard @Dantheman720 

I've got a PR in to fix this issue ( https://github.com/redhat-developer/developers.redhat.com/pull/1685 ). It looks like when the product pages were migrated it was using the old template that didn't just dynamically inject the alert. I took out the hard-coded element, so now we should only be getting one on the page. As soon as someone gets in to see this they can check and should be good to merge that PR.

RubyGemTSL:
retest this please

RubyGemTSL:
thanks @KamiQuasi - merged your PR

RubyGemTSL:
retest this please

PR #1696, opened 2017-01-13
Fixes for issues found during analytics training

KamiQuasi:
Fixes
- Download event firing without an actual download
- URL parsing to correct incorrect pageName and categories


LightGuard:
Merging anyway. We timed out.

PR #1697, opened 2017-01-14
Revert "Fixes for issues found during analytics training"

paulrobinson:
Reverts redhat-developer/developers.redhat.com#1696


LightGuard:
Checking if this is the issue with downloads or not.

LightGuard:
Yep, downloads work correctly on staging now. This is a pretty high priority issue, but I'll see if tests actually work.

paulrobinson:
Agreed, looks to be working. Need to see it working with the export first, and check it doesn't break tests

paulrobinson:
Export works too.

RubyGemTSL:
@LightGuard @paulrobinson - FYI tests would not have caught this as download tests are still disabled due to export hold issues. There was also changes to the registration page at the end of last week that broke the registration tests.

paulrobinson:
OK. I think we need to chase this up again. The lack of download tests is really hurting us

PR #1698, opened 2017-01-16
Fixed registration tests and re-enabled download tests.

RubyGemTSL:


RubyGemTSL:
@paulrobinson - even if we see export hold issues from time to time I think given the recent problems that have been missed by not testing downloads we should keep them enabled. We just need to be mindful that the kc_dm tests won't always pass and will need to check out the cucumber report when reviewing/merging. cc @LightGuard.

LightGuard:
Is this good to go?

RubyGemTSL:
@LightGuard yes please! Thanks!

RubyGemTSL:
@paulrobinson this PR includes hardcoded users to test the download flow. 80% of download tests are still being ignored due to Export Hold issues.

PR #1702, opened 2017-01-17
[DEVELOPER-3759] Removed deprecated environments from build scripts

robpblake:
This pull requests removes support for the `awestruct-dev` and `awestruct-pull-request` environments.

Additionally it now makes the `drupal-dev` environment the default when using the `control.rb` build script.


robpblake:
@LightGuard This should be good to go I think. If you have time to give it the once over today, I would appreciate that.

Maybe just have a quick play with it locally to ensure that switching the default environment to `drupal-dev` works as expected and doesn't cause any surprises.

cheers,

Rob

robpblake:
retest this please

LightGuard:
Worked for me!

PR #1710, opened 2017-01-19
Update Sentry.io raven script to latest version

KamiQuasi:
Latest version is 3.9.1


LightGuard:
@KamiQuasi Is there really anything to test here besides the JS works?

KamiQuasi:
@LightGuard not really anything to test besides that. We'll want to put the change onto the production config as well.

LightGuard:
Added to prod

PR #1711, opened 2017-01-20
[DEVELOPER-3500] Remove references to Drupal host from site export

robpblake:
This PR tries as much as possible to remove all references to the Drupal host from the export of the site.

I've updated the export process to do a few more things:

1. Prevent httrack from adding the "Mirrored from..." to the source of every HTML file. Ideally I wanted to make it add nothing to the file, but this doesn't work when updating a cache of an existing export. So I've configured it to add an empty HTML comment instead

2. It will remove any of the following HTML elements from the DOM of each HTML file in the export:

- `<link rel="shortcut">`
- `<link rel="revision">`
- `<meta name="Generator">`

I've ensured that we only re-write each HTML file once, so the export process needed slightly more tweaking than I originally hoped.

Please note that currently I cannot remove the reference to the Drupal host from the Javascript on the page. It just makes our life difficult to do this.

**How to verify:**

- Visit the exported version of the site associated with this PR
- Pick a few pages at random
- View the source for the page and note that the only reference to the Drupal host is now within JS on the page

Additionally please make sure that the search function still works on each page you visit as I had to slightly re-work the code that modifies the form target for search.

### DIFF
--- a/_docker/lib/export/README.md
+++ b/_docker/lib/export/README.md
@@ -31,7 +31,7 @@ The export process relies on [httrack](https://www.httrack.com/) to perform the 
 The default configuration of httrack is as follows:
 
 ```
---list /export/url-list.txt -O /export --disable-security-limits -c50 --max-rate 0 -v +"http://#{drupal_host}*" -"*/node*" -o0 -N ?html?%h/%p/%n/index.html -N %h/%p/%n.%t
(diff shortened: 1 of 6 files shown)


robpblake:
@paulrobinson Could you take a look at this one to make sure you're happy that we've done "enough" to close this issue for now. As mentioned it is going to cause us larger headaches to try and fix the JS right now.

cheers,

Rob

paulrobinson:
I think this is a step in the right direction, but I suspect we will need to remove the host from the JS to mark the issue as done. I'll check with Lukas.

PR #1712, opened 2017-01-21
Google Webmaster Tools authentication file

paulrobinson:
This file is needed by Jairus Mitchell in order to convince Google that we have authority to access Google Web master tools for developers.redhat.com.

This works if you see 'google-site-verification: googlead73883659545ab5.html' when browsing to https://developers.redhat.com/googlead73883659545ab5.html (or the equivalent when verifying the PR).

### DIFF
--- a/_docker/lib/export/static/googlead73883659545ab5.html
+++ b/_docker/lib/export/static/googlead73883659545ab5.html
@@ -1,0 +1,1 @@
+google-site-verification: googlead73883659545ab5.html
--- a/googlead73883659545ab5.html
+++ b/googlead73883659545ab5.html
@@ -1,1 +1,0 @@
-google-site-verification: googlead73883659545ab5.html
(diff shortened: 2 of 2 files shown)


robpblake:
@paulrobinson Rather than creating an empty page in Drupal (which we then have to export), can I suggest we simply drop this into the `static` directory of the export. Resources in that directory are copied "as-is" into the root of the export once it is finished.

You can find that directory in `_docker/lib/export/static`.

robpblake:
retest this please

paulrobinson:
Sure, I'll move the file. 

I don't think it needs to be in the sitemap.

paulrobinson:
@robpblake Done.

robpblake:
retest this please

robpblake:
retest this please

robpblake:
Checked export and the page is present. Checked the Drupal instance and the page is not there (as expected.)

PR #1713, opened 2017-01-23
DEVELOPER-3770 - Set registered flag based on rhd_member cookie

KamiQuasi:
https://issues.jboss.org/browse/DEVELOPER-3770

Sets the Adobe DTM registered flag on the user based on the existence of the rhd_member cookie.


KamiQuasi:
To test, if you are not logged in you should see "false" for this value in the developer console: digitalData.user[0].profile[0].profileInfo.registered

If you have logged in, either actively or previously, you should see "true" for the value.

KamiQuasi:
Unfortunately, it appears that since the cookie is set by the staging domain on our PR environment, so the cookie doesn't reflect the correct value because it can't be read across domains. I'll say that the code isn't breaking anything, and it is working properly, so this should be able to be merged.

PR #1716, opened 2017-01-23
DEVELOPER-3674 Created Connectors content type and REST export view

adelasofia:


adelasofia:
retest this please

LightGuard:
Some help text for the fields would be nice. Also not sure if we need a short description or if you could simply use the summary of a long description. I don't see any permissions setup for this content type either.

LightGuard:
Looks okay to me from a code/configuration perspective.

Dantheman720:
@adelasofia - Aside from the addition of data validation, everything looks good to me.

https://github.com/adelasofia/developers.redhat.com/blob/master/_docker/drupal/drupal-filesystem/web/modules/custom/rhd_common/rhd_common.module is what we have been use to add field validation.

LightGuard:
Looks good, let's see how tests go.

PR #1720, opened 2017-01-26
DEVELOPER-3752 - Fix double-call of search on resources

KamiQuasi:
https://issues.jboss.org/browse/DEVELOPER-3752

Added logic to prevent timeout call in addition to initialization call. If on resources page, just always restore filters as it will call the updateSearch function.

May also resolve - https://issues.jboss.org/browse/DEVELOPER-3691 due to two calls to the search service being made.

### DIFF
--- a/javascripts/search.js
+++ b/javascripts/search.js
@@ -752,7 +752,7 @@ function searchCtrlFunc($scope, $window, searchService) {
       window.location = '#!';
     }
     window.setTimeout($scope.filter.restore, 0);
+  } else {
+    $scope.updateSearch();
   }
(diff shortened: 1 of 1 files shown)


KamiQuasi:
@LightGuard or @Dantheman720 to verify this you'll probably want to snag the Omnibug Chrome Extension. If you load /resources on production you'll see two "Internal Search" click events at the end of the list. With this PR you'll only see one. https://chrome.google.com/webstore/detail/omnibug/bknpehncffejahipecakbfkomebjmokl?hl=en

PR #1724, opened 2017-01-27
Expanded the drupal rest export to display all content

Dantheman720:


LightGuard:
What's the reason for this?

Dantheman720:
In connectors, the view was set to only display 10 items at a time. This makes it a little bit more difficult to index with DCP because of pagination.

PR #1730, opened 2017-01-31
Fix for tests not running in mobile emulated browser when mobile prof…

RubyGemTSL:
…ile is set.


RubyGemTSL:
@paulrobinson can you please merge this for me? @robpblake is happy with my changes.

PR #1732, opened 2017-01-31
DEVELOPER-3646 - Topics header graphic missing on a few pages

Dantheman720:


_docker/drupal/drupal-filesystem/web/themes/custom/rhd/templates/topic/node--topic.html.twig
```diff
@@ -65,9 +65,13 @@ other methods (such as node.delete) will result in an exception.
 * @ingroup themeable
 */
 #}
-{% set topic_name = label[0]['#context']['value']|lower %}
+{% set topic_name = label[0]['#context']['value']|lower|replace(' ', '-') %}
```
LightGuard: Actually, the clean_class filter would probably be better. It eventually calls https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Utility%21Html.php/function/Html%3A%3AcleanCssIdentifier/8.2.x which you can see does a lot more than this does.

Dantheman720:
retest this please

LightGuard:
Fixes everything but .net, but I'm guessing we need to make a css change for that to take effect.

PR #1733, opened 2017-02-01
DEVELOPER-3812 - Investigate & fix random failing SO tests 

RubyGemTSL:
This PR is a fix for the random failing SO tests that sometimes occur on PR test runs.

Tests were failing when a scenario clicked on a link that takes the user to stackoverflow.com, the driver was closing the window before the the  stackoverflow.com tab was opened, therefore tests were bombing out.


RubyGemTSL:
Ready to be merged.

PR #1736, opened 2017-02-01
Allowing select and option in rhd approved html

LightGuard:
Please see https://issues.jboss.org/browse/DEVELOPER-3831


LightGuard:
Tiffany says this is good to go in. Please merge.

PR #1738, opened 2017-02-02
[DEVELOPER-3758] Prevent backup from using all disk space on production VM

robpblake:
This pull request introduces a couple of utility classes that help us reset the Git LFS repository checkout on the production VM.  The purpose for doing this is to prevent the Git repository consuming all available disk space on the VM.

The process here is pretty simple, it basically:

1. Deletes the current checkout of the rhd-backups repository
2. re-clones the repository using `git lfs clone`

`git lfs clone` is smart enough to only pull the files within the current branch i.e. master branch for a fresh checkout. Therefore we immediately reclaim the disk-space in use by the repository.

The `backup_reset.rb` class here is designed to run with the `backup` service Docker container definition in the `drupal-production` environment. It is designed this way so that credentials for the Git repo and access to the filesystem are managed as with all other jobs that run in that environment. We don't have to start storing credentials in Jenkins.

The `run.rb` is a small wrapper script that is executed by Jenkins that ensures we execute the correct `docker-compose` command to get the `backup_reset.rb` script to what it needs to do.

This is all tied together by the new job [developers.redhat.com-drupal-backup-reset](http://jenkins.mw.lab.eng.bos.redhat.com/hudson/view/jboss.org/job/developers.redhat.com-drupal-backup-reset/) in Jenkins. This job has been configured to run at 04:00 each Monday morning ensuring that each week we start with a fresh checkout of the backup repository.

### DIFF
--- a/_docker/lib/backup/reset/backup_reset.rb
+++ b/_docker/lib/backup/reset/backup_reset.rb
@@ -1,0 +1,48 @@
+require_relative '../../default_logger'
+require_relative '../../../lib/process_runner'
+
+#
+# This simple class performs a simple reset of the Git repository that we use to store backups in order to
+# reclaim disk space.
+#
+# @author rblake@redhat.com
+#
+class BackupReset
+
+  def initialize(backup_directory, git_backup_repository, process_runner)
+    @backup_directory = backup_directory
+    @git_backup_repository = git_backup_repository
+    @process_runner = process_runner
+    @log = DefaultLogger.logger
+  end
+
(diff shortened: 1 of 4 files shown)


robpblake:
@LightGuard / @Dantheman720 Can one of you please give this the once over to make sure it all makes sense.

I don't think there is an easy way to fully test this one, so unfortunately we may have to just rely on the unit tests I've added and a code review from one of you.

cheers,

Rob

PR #1741, opened 2017-02-03
[DEVELOPER-3768] Improving Drupal boot time and Awestruct performance

robpblake:
This PR does two things:

1. Significantly improves the boot time of Drupal in the drupal-dev environment
2. Significantly improves the performance of Awestruct in the drupal-pull-request environment

**1: Improving Drupal boot time in drupal-dev environment**

To improve the boot time of Drupal in the drupal-dev environment, I have significantly reduced the size of the database that gets imported from ~10G down to ~2G. To do this I have simply removed all content from the `node_revision__body` table in the `drupal` schema. Blunt, but this seems to work OK,  but obviously we lose all revision data in drupal-dev environment. 

To use the reduced DB , we now publish a `drupal-data-lite` Docker image. This image is generated by the following Jenkins job: http://jenkins.mw.lab.eng.bos.redhat.com/hudson/view/jboss.org/job/developers.redhat.com-drupal-data-container-lite/ This job runs as a downstream of the current data container generation job and essentially removes the `node_revsion__body` table content from the database dump in the `redhatdeveloper/drupal-data` image and then re-pushes the image into the `redhatdeveloper/drupal-data-lite` repository. The `drupal-dev` environment has been updated to use the `drupal-data-lite:latest` image when creating the `drupal-data` container instance.

Testing on my machine reduces the boot time of MariaDB (and hence Drupal) down to about 2 minutes from ~15 minutes prior to this.

In addition I have provided a my.cnf configuration in `drupal-dev` that attempts to apply recommended best practice to the MariaDB instance. This will have to be a bit of a suck-it-and-see approach as obviously, everyone has different sized machines. I've tried to be conservative, but at the same time, tune MariaDB per some Drupal recommendations.

**2. Improving performance of Drupal in pull requests**

Its important that we test against the full Drupal dataset and therefore we will continue to use the `redhatdevelopers/drupal-data` image in the `drupal-pull-request` environment and not the "lite" version.

I have instead here provided a much more aggressive MariaDB configuration that makes use of the abundance of resources we have on Stumpjumper. From initial testing this seems to reduce the time to run Awestruct from anywhere between 25-45 minutes, down to about 10 minutes.

**For Reviewers**

Please do the following:

1. `--run-the-stack` locally on your machine and ensure Drupal boots as expected
2. Have a play around with Drupal ensuring that all of the content is there, including content only authored in the production Drupal instance.


LightGuard:
What does this bring to the table for us? We can't use this configuration in production.

robpblake:
@LightGuard It reduces the run-time of Awestruct in the PR environment from anywhere between 25-45 minutes to about 10 minutes. 

There is no intention here to use this in production, more an attempt to reduce the build times for each pull request.

cheers,

Rob

robpblake:
retest this please

robpblake:
retest this please

robpblake:
retest this please

robpblake:
@LightGuard Could you please have a play with this one locally. Dropping just the content of `node_version__body` seems to leave Drupal working OK, but I'd definitely like a 2nd set of eyes on it.

cheers,

Rob

robpblake:
@KamiQuasi Could you please have a play with this one on your machine as a fellow Mac user. It seems to work OK for me and I'd be interested to see if there is much in the way of performance improvement on your machine.

cheers,

Rob

LightGuard:
Here are my results running this:

Full run: 20:43.40
"primed run": 15:46.17 (Don't have to rebuild docker images.)
No awestruct: 3:52.95

I feel the experience before took about double that time.

All of these are using the same DB. Drupal seems to be functioning correctly as well. I'd love to hear about Luke's experience using this PR. From my perspective, I'm good with this.

PR #1742, opened 2017-02-06
DEVELOPER-3697 Events page css incorporation

adelasofia:


adelasofia:
There is some CSS inline in /events. This PR adds these styles to events.scss and once merged I'll remove the inline styles from the page.

PR #1748, opened 2017-02-08
DEVELOPER-3817 - Add drop-down for language on RHEL

KamiQuasi:
https://issues.jboss.org/browse/DEVELOPER-3817

Adds the ability to use Angular attributes on div, a, and select tags to allow the changing of language values to hide/show content accordingly.


_docker/drupal/drupal-filesystem/web/themes/custom/rhd/css/base/eclipse.css
```diff
@@ -0,0 +1 @@
+.wide.wide-hero.eclipse{background:url(/images/eclipse/eclipse_microsite_hero.jpg);background-size:cover;height:450px}
```
robpblake: @KamiQuasi Its worth checking with @LightGuard whether this should be committed. I believe most of the CSS in this directory is generated, so not sure if this should be here permanently.

cheers,

Rob

KamiQuasi: @robpblake I was wondering about that. It showed up after a build in my git status, but I thought it was generated because of the changes I made. @LightGuard do I just need to remove this and update the .gitignore?

LightGuard: I have no idea what generates this. I haven't looked into it either. I'd recommend removing it from the PR. If you'd like to take the extra time to figure out what's generating it, even better. First thoughts, it's something in the Gulp script.

PR #1750, opened 2017-02-10
[DEVELOPER-3552] Fix broken 404 error page in export

robpblake:
This PR fixes broken error pages on the site export from Drupal.

It does this by re-writing all absolute links on the error pages back to an absolute form. The value used to replace the relative form of the link is read from the environment variable `drupal.export.final_base_url`. If this value is not set, then it will default to `https://developers.redhat.com`. Any other environment than production should set this variable on the `export` container definition in the environment to ensure that 404 and other error pages are re-written as expected.

At the minute this change only gets applied to the `404-error` and `general-error` page.

**For reviewers:**

1. Navigate to the export of this PR
2. Ensure that `/404-error` page renders correctly, that there are no load errors in the console, and inspect the source to ensure that all links are absolute
3. Do the same for the `/general-error` page

### DIFF
--- a/_docker/environments/drupal-dev/docker-compose.yml
+++ b/_docker/environments/drupal-dev/docker-compose.yml
@@ -99,6 +99,7 @@ services:
     entrypoint: "ruby _docker/lib/export/export.rb docker"
     environment:
      - drupal.export.fail_on_missing
+     - drupal.export.final_base_url=http://docker:9000/
 
   #
   # Testing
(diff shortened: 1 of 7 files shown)


robpblake:
@paulrobinson / @KamiQuasi Can you please both take a look at this one to make sure its working as expected.

cheers,

Rob

robpblake:
retest this please

robpblake:
retest this please

LightGuard:
@robpblake Looks good, but https://developers-pr.stage.redhat.com/pr/1750/export/products/bleh/foo.html gives me a generic apache 404. Is that expect?

robpblake:
@LightGuard Yes. The configuration for the pull request environment Apache instance does not re-direct you to the correct 404 location.

PR #1774, opened 2017-02-27
DEVELOPER-3801 - test fixes

RubyGemTSL:
Fixed non passing/failing test. .present? returns true or false. Added an assertion, and to keep things consistent added footer element to site base.


RubyGemTSL:
@adelasofia FYI .present? returns true or false. Always remember to assert an element is present. Apologies I should have put that in the README.md. I will update. But if you review this test it is only a minor change, and you weren't to know.

robpblake:
retest this please

PR #1778, opened 2017-03-01
Local Drupal setup scripts

LightGuard:
Updates to the local development setup and the readme.

I think most people have tried this, but I'd appreciate you all trying this new version. I've also done some improvements to the README, please look at those too.


visuale:
I got an odd error while running `brew install php70 php70_xdebug composer`

`Error: No available formula with the name "php70_xdebug" `

visuale:
I tried this a bit more... getting the site functioning but is it supposed to remain at the 127.0.0.1:888 URL root? When viewing the site at that address, these two files (by extension styles and JS) turn up missing...

- http://127.0.0.1:8888/themes/custom/rhd/css/base/app.css?omd6kl 

- http://127.0.0.1:8888/themes/custom/rhd/js/all.min.js?omd6kl

Clicking sub-topic pages, however, changes the numeric URL to this URL root  http://developer-drupal.web.prod.ext.phx2.redhat.com, which shows everything correctly. Is that the expected result?

PR #1781, opened 2017-03-04
DEVELOPER-3678 - Stubbed resources tests.

RubyGemTSL:
DEVELOPER-3678 - Stubbed resources tests.

This PR should fix the issues with stubbing the resources page. I found that navigating directly to /resources does not redirect the user to https (using phantomjs). I modified the tests to navigate to resources via the primary nav menu. This way the (expected) access-control-origin of `https://developers-pr.stage.redhat.com` used rather than access-control-origin of `http://developers-pr.stage.redhat.com` which I think is the reason we're not hitting the stubbed data.


RubyGemTSL:
@robpblake could you please review/merge this today? Thanks!

PR #1787, opened 2017-03-07
[HOLD] DEVELOPER-3862 Updated user Auth header

adelasofia:
Jira: [DEVELOPER-3862](https://issues.jboss.org/browse/DEVELOPER-3862)

Progressive profiling support for RHD user management brings situation when user may be created without first and last name. RHD Website SSO code should show username instead of full name in the header if full name is not available. Check content of name field obtained from SSO token, if it is empty or blank (only space character) then it should show username instead in app.sso javascript function.


adelasofia:
This has been tested and approved by Vlastimil E.

PR #1790, opened 2017-03-08
DEVELOPER-3935 - Remove phantomjs and chromedriver binaries

RubyGemTSL:
This PR removes the need to have locally stored binary files in the acceptance test repo.


RubyGemTSL:
@robpblake should be good to merge if your happy with it. Thanks!

PR #1796, opened 2017-03-13
DEVELOPER-3922 - Add referrer policy for proper download forwarding

KamiQuasi:
https://issues.jboss.org/browse/DEVELOPER-3922

We may want to rethink this a little before merging. There could be unintended security side-effects of enabling the "unsafe-url" option. This does not seem to be able to be set per environment since it is a Drupal configuration value. Need feedback from @dhladky to see if a less liberal option like "strict-origin-when-cross-origin" would work for this particular need or if that would still prevent proper forwarding after a download.

References:

https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-unsafe-url
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy


LightGuard:
@KamiQuasi, you could look into https://api.drupal.org/api/drupal/core%21includes%21theme.inc/function/template_preprocess_html/8.2.x and the example on that page about modifying the head-placeholder for the meta stuff and do it per environment.

robpblake:
retest this please

robpblake:
retest this please

LightGuard:
@KamiQuasi Are you good with this being merged?

KamiQuasi:
retest this please

LightGuard:
Where are we with this?

KamiQuasi:
At the moment, it appears that @dhladky may have some changes to do on the download manager, but @RubyGemTSL may have other thoughts. This fix, as it is, can probably be merged since it does what @dhladky had requested, but the redirect after a download appears to add on an extra "download" to the path back to the PR server which causes the resulting page to be a 404.

LightGuard:
We're not seeing the redirect 404 currently, correct? Is it an environmental thing we'll only see in PR?

KamiQuasi:
@LightGuard yeah, this was all to fix downloads on PR environments so that they would redirect back to the PR server rather than Staging. It appears they're at least redirecting to the PR server, but the address is incorrect.

LightGuard:
@KamiQuasi great, so this won't affect production at all. At least that's what I'm understanding (and what the code looks like).

KamiQuasi:
@LightGuard correct, this should only ever apply on non-production environments.

PR #1800, opened 2017-03-16
Developer 3945 - Update Registration tests

RubyGemTSL:
This PR fixes the broken registration flow for KC DM tests. Until recently the flow was broken, now it has been fixed and verify email address has been introduced.


RubyGemTSL:
@robpblake registration flow has finally been fixed! Can you get this merged today then we can finally have green PR's!

PR #1804, opened 2017-03-21
DEVELOPER-3948 setting up Drupal Unit Tests

LightGuard:
We probably need to break up
https://issues.jboss.org/browse/DEVELOPER-3948 but here's a stab at
getting the infrastructure in place.

Unit tests for the rhd_common module can now be run via our control
script. We probably need to make some changes to the jobs in order to
get these to run as part our builds.

Drupal Unit Tests can be run using either `-d` or `--drupal-unit-test`.
This is probably the easiest way to run tests as running them locally
involves some knowledge of how to run it all.

Rob and I agreed that we should use the current infrastructure and
practices for functional and end-to-end testing.

### DIFF
--- a/_docker/control.rb
+++ b/_docker/control.rb
@@ -373,6 +373,10 @@ if $0 == __FILE__
     system_exec.execute_docker_compose(environment, :run, tasks[:unit_tests])
   end
 
+  if tasks[:drupal_unit_tests]
+    system_exec.execute_docker_compose(environment, :run, tasks[:drupal_unit_tests])
+  end
+
   start_and_wait_for_supporting_services(environment, tasks[:supporting_services], system_exec)
 
   if tasks[:awestruct_command_args]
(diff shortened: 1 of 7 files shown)


robpblake:
@LightGuard Some comments:

1. Is there a requirement to have the results of the unit testing reported as a status on the GitHub pull request?
2. It would be nice if the Jenkins job was able to archive the XML output of running the tests. To that end, we would require a volume in the container that runs the phpunit tests so that we can write the XML file to the Jenkins workspace
3. Something I've been thinking about is why we run the tests so much with the control.rb script. Having the tests run as part of the `--run-the-stack` option forces the declaration of the unit_tests/drupal_unit_tests services in environments where they are not really required e.g. There really shouldn't be any need to run either of the unit test suites in production or staging environments as code should not enter those environments if the unit tests have failed. Is there any objection to the tests running only with the `-d` option and not as part of `--run-the-stack`?

LightGuard:
@robpblake:

1. Actually worked backward on this one. We currently don't have any requirements, this actually started more exploratory than anything, but turned out it wasn't very difficult. We could certainly add that if we'd like.

2. The xml file is saved out to `/tmp/test_output.xml` currently. We could just as easily save that somewhere else and pull it out into the workspace via a mount or `docker cp`.

3. I have no problems with that.

PR #1806, opened 2017-03-22
DEVELOPER-3844 Created cheat-sheet content type

adelasofia:
[DEVELOPER-3844](https://issues.jboss.org/browse/DEVELOPER-3844) - Created "Cheat Sheets" Content Type

The Cheat Sheet instances will be present under the https://develpers.redhat.com/cheat-sheet/<name>" URL.

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/config/sync/field.storage.node.field_cheat_sheet_image.yml
+++ b/_docker/drupal/drupal-filesystem/web/config/sync/field.storage.node.field_cheat_sheet_image.yml
@@ -1,0 +1,30 @@
+uuid: d9d91bcf-da23-4f80-abbd-ac5f13df3bc9
+langcode: en
+status: true
+dependencies:
+  module:
+    - file
+    - image
+    - node
+id: node.field_cheat_sheet_image
+field_name: field_cheat_sheet_image
+entity_type: node
+type: image
+settings:
+  uri_scheme: public
+  default_image:
+    uuid: ''
+    alt: ''
+    title: ''
+    width: null
+    height: null
+  target_type: file
+  display_field: false
+  display_default: false
+module: image
+locked: false
+cardinality: 1
+translatable: true
+indexes: {  }
+persist_with_no_fields: false
+custom_storage: false
--- a/_docker/drupal/drupal-filesystem/web/themes/custom/rhd/templates/node--cheat-sheet.html.twig
+++ b/_docker/drupal/drupal-filesystem/web/themes/custom/rhd/templates/node--cheat-sheet.html.twig
@@ -105,7 +105,7 @@ other methods (such as node.delete) will result in an exception.
             {{ content.field_cheat_sheet_image }}
         {% endif %} #}
         <a  class="button hidden-after-login" href="{{ content.field_cheat_sheet_download_url[0]['#url'] }}">Get The Cheat Sheet</a>
-        <a  class="button shown-before-login" href="{{ content.field_cheat_sheet_download_url[0]['#url'] }}">Download</a>
+        <a  class="button shown-after-login" href="{{ content.field_cheat_sheet_download_url[0]['#url'] }}">Download</a>
     </div>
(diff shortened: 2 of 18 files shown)


adelasofia:
This has been approved by Tiffany

adelasofia:
retest this please

RubyGemTSL:
Looks like there is an issue with registrations... again. Will investigate. Don't hold back on merging this as it has nothing to do with registrations.

_docker/drupal/drupal-filesystem/web/themes/custom/rhd/templates/node--cheat-sheet.html.twig
```diff
@@ -0,0 +1,111 @@
...
+                </li>
+            </ul>
+        </div>
+    </div>
+  </div>
+  <div class="row">
+    <div{{ content_attributes.addClass(["large-18", "medium-15", "columns"]) }}>
+        {{ content.body }}
+    </div>
+    <div class="large-6 medium-5 columns">
```
LightGuard: This doesn't add up to 24 for medium

_docker/drupal/drupal-filesystem/web/themes/custom/rhd/templates/node--cheat-sheet.html.twig
```diff
@@ -0,0 +1,111 @@
...
+        {% if content.field_cheat_sheet_tags|render is not empty %}
+            <strong>Tags:</strong> {{ content.field_cheat_sheet_tags }}
+        {% endif %}
+    </div>
+    <div {{ content_attributes.addClass(["large-20", "medium-15", "columns"]) }}>
```
LightGuard: Is this intended to be a new row? If so, we should setup the div to be within a new row div.

stylesheets/app.scss
```diff
@@ -201,10 +202,9 @@ button, .button {
     margin-left: 20px;
   }
   &.blue {
-    background:$med-blue;
-    &:hover {
-      background: lighten($med-blue,10%);
-    }
+    background:white;
```
LightGuard: This is going to affect ALL buttons and links that look like buttons, is that the intended effect?

_docker/drupal/drupal-filesystem/web/config/sync/field.storage.node.field_cheat_sheet_image.yml
```diff
@@ -0,0 +1,30 @@
+uuid: d9d91bcf-da23-4f80-abbd-ac5f13df3bc9
```
LightGuard: We may need to modify this in the future due to the issue of images disappearing.

stylesheets/app.scss
adelasofia: Yes, this would affect all buttons with the CSS class 'blue'. But there's actually no current buttons with that class on site (only on jboss.org pages that were leftover but no longer used). 

UX removed the button this change was made for on the last moment of the review process. But I still left the changes for future buttons that have this look, which is the right look for 'blue' buttons according to VD.

_docker/drupal/drupal-filesystem/web/themes/custom/rhd/templates/node--cheat-sheet.html.twig
adelasofia: This is not intender to be in a new row, it is inside the row in line 90

_docker/drupal/drupal-filesystem/web/themes/custom/rhd/templates/node--cheat-sheet.html.twig
adelasofia: Just pushed a fix for this 👍

stylesheets/app.scss
KamiQuasi: Sounds good to me. Might as well leave it in even if it isn't impacting this particular PR.

PR #1816, opened 2017-03-31
Sentry Update - Quote environment string

KamiQuasi:
Environment string needs to be quoted.


LightGuard:
Good to merge once everything passes.

PR #1820, opened 2017-04-04
[DEVELOPER-3985] Fix generation of drupal-data-lite Docker image

robpblake:
This PR introduces a fix for the generation of the drupal-data-lite Docker image. The current generation of this image is failing due to lack of space within the container filesystem. As the size of the production database grows, this will continue to be an issue. Therefore this PR provides a permanent fix to the issue.

It makes use of the `builder` pattern with Docker to generate the "lite" database `.sql.gz` to a volume rather than the local container filesystem. Once the "lite" data dump has been generated, it is then added to a Docker Image, and the resulting image is pushed to DockerHub.

Additionally the job within Jenkins has been updated to execute this new approach.

**To test**

I have tested this manually myself, but to test you can:

`ruby _docker/drupal-data/run_lite.rb`

This will build the new image and push it to DockerHub. A word of warning: The push can take a while if you don't have a particularly large upstream amount of bandwidth.


robpblake:
@LightGuard Would you mind just casting your eye over the code on this one. It will permanently fix the issues we are having with the generation of the `drupal-data-lite` image.

I have already tested it manually myself, but if you want to test it locally, details of how to do that are in the PR description.

cheers,

Rob

PR #1834, opened 2017-04-24
DEVELOPER-4053 Added WebpageSearchExclude meta tag

adelasofia:
[DEVELOPER-4053](https://issues.jboss.org/browse/DEVELOPER-4053)

Add `<meta name="DCP:WebpageSearchExclude" content="true">` to webpages that should not appear in the rht_website results. I.e those that have a more specific version of the content.

### DIFF
--- a/events/devnation/2016/index.html.slim
+++ b/events/devnation/2016/index.html.slim
@@ -111,5 +111,8 @@ section.wide.wide-hero(class="#{page.hero_class}")
                   - if c.speaker
                     p #{c.speaker_intro}
 
+noscript
+  meta name="DCP:WebpageSearchExclude" content="true"
+
 = javascripts('devnation2016-index', true) do
   script src="#{site.base_url}/javascripts/devnation2016.js"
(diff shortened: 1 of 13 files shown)


adelasofia:
@unibrew Can you take a look to see if this works for you?

adelasofia:
retest this please

events/devnation/2016/index.html.slim
```diff
@@ -6,7 +6,6 @@ interpolate: true
 hero_class: devnation2016
 title: Devnation 2016
 description: Learn more. Code more. Share more. Join the Nation.
-ignore_export: true
```
LightGuard: Why are we removing this?

adelasofia:
@LightGuard You're right, I just added back the `ignore_export: true` to the layouts, but I left the meta tags in case they are ever pushed into drupal again. I'll add the meta tags dirctly to drupal for those pages.

LightGuard:
Looking good.

LightGuard:
retest this please

unibrew:
I took a look at https://developers.redhat.com/video/youtube/rxPctRaTznc/ why this meta tag is somewhere deep inside the <body> instead of <head> ?

LightGuard:
Because adding custom meta tags in Drupal is nontrivial. This was a much easier solution and actually still valid HTML oddly enough.

PR #1835, opened 2017-04-24
DEVELOPER-4028 - Update RHD logo with the RHDP logo

KamiQuasi:
https://issues.jboss.org/browse/DEVELOPER-4028


KamiQuasi:
@adelasofia I didn't see any issues with position or size when I dropped the window and hit any breakpoints, but take a look and see if you see anything off.

PR #1836, opened 2017-04-24
DEVELOPER-3976 CKEditor buttons for media library

LightGuard:
JIRA: https://issues.jboss.org/browse/DEVELOPER-3976

Within CKEditor (the WYSIWYG editor in Drupal) you should see two new icons: a star and an 'E'. The star is for media (there may be two already in the view). The 'E' is for embedding nodes.

You should be able to click on the star, upload a new image into the media library and place an image from the library into the editor.


LightGuard:
retest this please

PR #1847, opened 2017-04-28
DEVELOPER-4056 Fixed articles layout nav

adelasofia:
[DEVELOPER-4056](https://issues.jboss.org/browse/DEVELOPER-4056)

Fixed: Side navigation in Article pages runs over footer when the height of the window is not tall enough to fit it in the content area.


LightGuard:
This looks to fix the problem, however, I noticed on mobile, the articles do not render correctly. The share buttons are on top of the content.

PR #1851, opened 2017-05-02
[FIX] - Update Sentry, move custom event call

KamiQuasi:
Updates PR Sentry environment to the most recent build and removes a dependency on DTM for some scripts to work by making the custom event sender global regardless of DTM being active.

### DIFF
--- a/_docker/environments/drupal-pull-request/rhd.settings.yml.erb
+++ b/_docker/environments/drupal-pull-request/rhd.settings.yml.erb
@@ -19,9 +19,9 @@ database:
   username: 'drupal'
   password: 'drupal'
   name: 'drupal'
-dtm_code: https://dpal-itmarketing.itos.redhat.com/developers
+dtm_code: 'https://dpal-itmarketing.itos.redhat.com/developers'
 rhd_base_url: "stumpjumper.lab4.eng.bos.redhat.com:<%= ENV['DRUPAL_HOST_PORT'] %>"
 rhd_final_base_url: "developers-pr.stage.redhat.com/pr/<%= ENV['ghprbPullId'] %>/export"
 sentry_track: true
-sentry_script: 'https://cdn.ravenjs.com/3.14.2/raven.min.js'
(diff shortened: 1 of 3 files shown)


RubyGemTSL:
retest this please

LightGuard:
Any reason not to merge this? @KamiQuasi do we need something done on Prod and staging as well?

KamiQuasi:
@LightGuard a couple changes for Prod and one for staging. Both need the Sentry version updated to 3.14.2 and then Prod needs the URL changed to https://c3d1750237a549fe91ac70c27793181b@sentry.io/69051

KamiQuasi:
@LightGuard change of version and adding a whitelist as I'm seeing some GTM or Adobe caused errors that we don't need to track. Once it gets through version for prod and stage will be 3.15.0, but the URL for Prod will be what I mentioned above.

KamiQuasi:
retest this please

_docker/environments/drupal-pull-request/rhd.settings.yml.erb
```diff
@@ -19,9 +19,9 @@ database:
   username: 'drupal'
   password: 'drupal'
   name: 'drupal'
-dtm_code: https://dpal-itmarketing.itos.redhat.com/developers
+dtm_code: 'https://dpal-itmarketing.itos.redhat.com/developers'
 rhd_base_url: "stumpjumper.lab4.eng.bos.redhat.com:<%= ENV['DRUPAL_HOST_PORT'] %>"
 rhd_final_base_url: "developers-pr.stage.redhat.com/pr/<%= ENV['ghprbPullId'] %>/export"
 sentry_track: true
-sentry_script: 'https://cdn.ravenjs.com/3.12.1/raven.min.js'
+sentry_script: 'https://cdn.ravenjs.com/3.15.0/raven.min.js'
```
LightGuard: This doesn't match what I just put in prod... should they?

LightGuard:
retest this please

PR #1854, opened 2017-05-09
[DEVELOPER-4023] Ensure sitemap.xml contains correct URL for each environment

robpblake:
This PR updates the fetch, save and re-write of the sitemap.xml during the export process to take into account the `drupal.export.final_base_url` environment parameter.

This allows the user to set the URL that should appear within the exported sitemap.xml on a per-environment basis. We already use this parameter to re-write URL locations on error pages as part of the export HTML post-processing, so integration here was fairly trivial.

For reviewers:

1. Allow the export for this PR to run
2. Visit the sitemap.xml for this PR and ensure that it has the correct URL for the export location and not https://developers.redhat.com
3. Ensure that the links within the sitemap.xml are pointing to genuine locations.

### DIFF
--- a/_docker/lib/export/drupal_page_url_list_generator.rb
+++ b/_docker/lib/export/drupal_page_url_list_generator.rb
@@ -3,6 +3,7 @@ require 'net/http'
 require 'nokogiri'
 
 require_relative '../default_logger'
+require_relative 'export_urls'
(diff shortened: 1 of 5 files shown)


LightGuard:
Looks good to me.

PR #1861, opened 2017-05-11
[DEVELOPER-4077] Lock npm dependency versions

robpblake:
This PR locks the versions of the npm dependencies used by the project to prevent using newer dependency versions that may not work.


robpblake:
@LightGuard This simply locks the npm dependency versions rather than relying on the `^` (compatible with). I don't see that this will cause any problems, but wanted to double check with you before merging as to whether there was a particular reason for the use of `^` in the first place?

If not and you're happy with this change, please go ahead and merge.

cheers,

Rob

PR #1867, opened 2017-05-15
DEVELOPER-4071 - Remove the requirement for an ISBN for books

Dantheman720:
This fixes the ISBN requirement. The export view should display an empty string if there is no ISBN available.


LightGuard:
I don't see any problems with the diffs.

Dantheman720:
Yeah, should be good to go.

PR #1868, opened 2017-05-16
Adding back in WYSIWYG editor video button

LightGuard:
Also making the rhd_html the format for all non-admin users.


LightGuard:
Go in and edit something and make sure the video button is back

LightGuard:
DM test failures, seeing as how this is a purely Drupal based change, I don't see any reason this would cause problems with the download manager.

adelasofia:
I see the video button 👍

PR #1870, opened 2017-05-16
DEVELOPER-4022 - Throw exceptions if empty/null

LightGuard:
JIRA: https://issues.jboss.org/browse/DEVELOPER-4022

If we don't have any revisions for a product or don't have an
active_paragraph/subpage throw a NotFoundHttpException. This will stop
the PHP errors and send a 404 back to the user, which is probably what
we want to have happen anyway.

The easiest way to test this would be to pull it in locally and run drupal, then go to a product that doesn't exist and a subpage that doesn't exist in an actual product (and for completeness a fake subpage in a product that is also fake). You'd then look for exceptions in the running HTTP server.


LightGuard:
Looks like all the failures are for KC or timeouts related to KC.

LightGuard:
Any chance we could get this reviewed and merged, please?

robpblake:
@LightGuard I've just tested this using your PR Drupal instance using the instructions you mentioned above. I see no errors in the Apache error_log and get a 404 page when:

- Browsing to a product page for a non-existing product
- Browsing to a non-existing sub-page of an existing product
- Browsing to a non-existing sub-page of a non-existing product

Looks like it's good to go.

PR #1899, opened 2017-06-15
[DEVELOPER-4212] Fixed image paths within _alerts.scss 

robpblake:
This PR fixes the image paths with _alerts.scss which is causing Awestruct to fail in staging and production.

Unfortunately I don't believe there is anyway to test this not in staging or prod as we don't use the cdn() function in PR environments.

### DIFF
--- a/stylesheets/_alerts.scss
+++ b/stylesheets/_alerts.scss
@@ -8,7 +8,7 @@
     border-color: #87aac1;
 
     &.icon {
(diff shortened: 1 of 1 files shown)


robpblake:
@KamiQuasi As far as I can tell this is good to go. It should resolve the issues with Awestruct failing in staging and prod.

cheers,

Rob

PR #1901, opened 2017-06-15
DEVELOPER-3946 - Create developers.redhat.com/login link and redirect

KamiQuasi:
Adds a script to redirect to login for any URL with /login

### DIFF
--- a/javascripts/sso.js
+++ b/javascripts/sso.js
@@ -4,7 +4,7 @@ app.sso = function () {
         var usr = digitalData.user[0].profile[0].profileInfo;
 
         if (window.location.href.indexOf('/login') >= 0) {
-            keycloak.login();
+            keycloak.login({"redirectUri": app.ssoConfig.logout_url});
         }
 
         if (keycloak.authenticated) {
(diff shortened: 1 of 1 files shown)


KamiQuasi:
retest this please

RubyGemTSL:
Before this get's merged. As Luke & I discussed on HipChat it may be a good idea to have something display something like 'Redirecting...' instead of the 404 page briefly.

KamiQuasi:
@RubyGemTSL when this is merged I'll do what I did on my local and put a new page at /login so the 404 doesn't flash. I've put one on the Drupal instance here for the sake of seeing how it works.
http://stumpjumper.lab4.eng.bos.redhat.com:36901/login

RubyGemTSL:
@KamiQuasi looks good to me.

Just a quick one, for testing purposes. Should be be able to go to https://developers-pr.stage.redhat.com/pr/1901/export/login? as I get a 404.

KamiQuasi:
@RubyGemTSL I don't know if the export re-runs when a change is made to the PR Drupal. I imagine it does not, but that would be why the export login page 404s.

PR #1904, opened 2017-06-15
[URGENT] DEVELOPER-4214 - Product download issue messaging

KamiQuasi:
https://issues.jboss.org/browse/DEVELOPER-4214

Puts info message on product and download pages.


paulrobinson:
We might have the download issue fixed soon., so lets hold on this for a moment.

paulrobinson:
Please link to the branch in the JIRA though, as we may need to do this kind of thing again one day.

paulrobinson:
The problem is fixed, so we no longer need this. Please do keep the code around, as this will likely form part of our response for future issues.

PR #1909, opened 2017-06-16
DEVELOPER-4075 - Both Topics and Technologies menu items are active for dotnet

KamiQuasi:
https://issues.jboss.org/browse/DEVELOPER-4075

When on dotnet topic and dotnet product pages only the correct navigation heading should be highlighted not both.


robpblake:
@KamiQuasi Good to merge. Checked the change and it looks to be working fine.

cheers,

Rob

PR #1927, opened 2017-06-23
Fix for Topics acceptance tests

adelasofia:


LightGuard:
Not sure what's happening here, but the tests are failing for redirects to the topic pages.

adelasofia:
retest this please

adelasofia:
@KamiQuasi I haven't been able to get these refferer tests to pass. Do you know what else needs updating?

KamiQuasi:
@adelasofia that was a tricky one. Take out the leading "/" on the feature's URL. The referrer check is kind of unforgiving, but what the test is navigating to ends up being //topics/devops and that doesn't work.

PR #1929, opened 2017-06-26
DEVELOPER-3994 - rel="canonical" absolute links

LightGuard:
Look at any page within Drupal and also some product pages. Make sure they are absolute links. Ideally the export process works just fine, however, they may not. This is just on the Drupal side for testing.


LightGuard:
@robpblake looks like we'll need a story to fix the export on this now.

PR #1932, opened 2017-06-28
DEVELOPER-4252 - Broken download page layout in Firefox

KamiQuasi:
https://issues.jboss.org/browse/DEVELOPER-4252

Specifically the /products/amq/download/ page layout in Firefox was all messed up.
Local environment isn't working at the moment, so putting this change in to give it a look in FF.


adelasofia:
@KamiQuasi the layout issue looks fixed, but I don't see the download tables in Chrome that I do see in safari and firefox

PR #1946, opened 2017-07-13
DEVELOPER-4246 - Retargeting event listener

KamiQuasi:
Retargeting the event listener from window to the component


KamiQuasi:
retest this please

PR #1954, opened 2017-07-18
DPAL updates for dev DTM

KamiQuasi:
Maintenance updates for dev environments to point to the correct URL for DTM staging since ITOS is being shut down.

@LightGuard or @paulrobinson this will need to be done on staging too. If you would change the Staging config's DTM code to point to: https://dpal-it-marketing.int.open.paas.redhat.com/developers

Also, if you want to update the Sentry config to point to version 3.17.0 this is probably a good time to do that too.

### DIFF
--- a/_docker/environments/drupal-dev/rhd.settings.yml
+++ b/_docker/environments/drupal-dev/rhd.settings.yml
@@ -24,7 +24,7 @@ database:
   username: 'drupal'
   password: 'drupal'
   name: 'drupal'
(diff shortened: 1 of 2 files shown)


robpblake:
@KamiQuasi Does the update to the dtm code endpoint and sentry version also need to be applied in production?

cheers,

Rob

KamiQuasi:
@robpblake the DTM update should not be done on production, but the sentry version bump should be done on production.

PR #1960, opened 2017-07-20
DEVELOPER-4245 Added loading image in search results

adelasofia:
[DEVELOPER-4245 - Add loading image while waiting for results](https://issues.jboss.org/browse/DEVELOPER-4245)

Requirements:
The amount of time to return search results can be multiple seconds. There should be an image showing that loading is happening so users do not get confused.

### DIFF
--- a/javascripts/build.js
+++ b/javascripts/build.js
@@ -523,6 +523,7 @@ var RHDPSearchApp = (function (_super) {
             this.results.more = e.detail.results;
         }
         this.count.count = e.detail.results.hits.total;
+        this.results.classList.remove('loading');
     };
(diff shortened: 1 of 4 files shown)


robpblake:
retest this please

RubyGemTSL:
@KamiQuasi @adelasofia I don't see the loading spinner on the site preview, do you?

KamiQuasi:
@RubyGemTSL I'm working on some comments for the PR, but yeah because of the way we're loading records I don't think the class change is showing up during the actual data load.

RubyGemTSL:
Thanks Luke. OK i'll hang fire on testing until it's ready. Sorry for treading on your toes :)

PR #1961, opened 2017-07-20
[URGENT] - DevNation Live Date check fix

KamiQuasi:


robpblake:
@KamiQuasi Is this one good to go? How should we test it?

cheers,

Rob

KamiQuasi:
@robpblake sorry about that (again). This one fixes a couple issues. First, the previous events should now show up on the right, and second the "Next" session should stay there until an hour after start time.

PR #1969, opened 2017-07-26
Updating the download link for OpenShift

LightGuard:
As per Erik Jacobs, the download link for OpenShift needs to be updated


KamiQuasi:
retest this please

PR #1971, opened 2017-07-26
DEVELOPER-4012 Books can now have their own page

LightGuard:
JIRA: https://issues.jboss.org/browse/DEVELOPER-4012

All the review steps are inside the JIRA.


KamiQuasi:
@LightGuard couple things I noticed:
* Got a message that "Book Excerpt" and "PDF Link" were required, but they don't appear to be in the interface. 
* Published date is not required, but "Published:" shows up orphaned on the page if it is not populated

PR #1980, opened 2017-08-04
DEVELOPER-4259 - Updated the view to display in EST instead of UTC

Dantheman720:


LightGuard:
Wouldn't it be better to send it as UTC and fix the display to use the timezone of the client?

robpblake:
retest this please

Dantheman720:
@LightGuard @tiffanynolan - I think that makes more sense. The only issue is that the client timezone will show up differently between the search page and the click through to the cheatsheets. The twig template picks up the time set by Drupal (static time) and the JS converts the timezone relative to the timezone of the browser (AFAIK). I could also explicitly state the time zone on the actual cheat sheet page.

For example: https://developers.redhat.com/cheat-sheets/red-hat-mobile-application-platform/

I can add UTC to the end of "Published: Thu, 07/06/2017 - 14:18"

LightGuard:
Of course Tiffany will trump me on this, but we should make these consistent, or state the timezone. 
		
		

		Jason Porter 
Sent from Outlook on iOS
	




On Mon, Aug 7, 2017 at 4:46 PM -0600, "Dantheman720" <notifications@github.com> wrote:












@LightGuard @tiffanynolan - I think that makes more sense. The only issue is that the client timezone will show up differently between the search page and the click through to the cheatsheets. The twig template picks up the time set by Drupal (static time) and the JS converts the timezone relative to the timezone of the browser (AFAIK). I could also explicitly state the time zone on the actual cheat sheet page.


For example: https://developers.redhat.com/cheat-sheets/red-hat-mobile-application-platform/


I can add UTC to the end of "Published: Thu, 07/06/2017 - 14:18"

LightGuard:
Are we waiting to see what @tiffanynolan has to say?

Dantheman720:
@LightGuard - That probably makes the most sense. At least the dates aren't showing January 18, 1970 anymore. They're only off by one because of the server client differences.

tiffanynolan:
@Dantheman720 @LightGuard What's the easiest way for these to show the same exact date? I don't think we need to show the time, just the **month day, year** would be great. Yes, it's way better than 1970 but it's still a very odd experience to see one date in the results and another on the page. I have no preference for which one you use, but they should be the same.

PR #2003, opened 2017-08-22
Add .NET promo to #1 and move OSIO to #3.

insectengine:


RubyGemTSL:
this is OK to test

LightGuard:
Are we good to merge, or do we need to wait?

PR #2007, opened 2017-08-24
Adding some more files to not push

LightGuard:
We'll need to fix the broken links in Drupal once this goes through.


KamiQuasi:
retest this please

PR #2021, opened 2017-09-01
DockerFilesLinkUpdate - Fixed broken links on jboss-docker

Dantheman720:


LightGuard:
I don't think we're actually pushing the page this would affect into Drupal anymore. Please double check.

Dantheman720:
@LightGuard - Ah! You are correct.

PR #2026, opened 2017-09-06
Upgrade Ruby for builds to pass

LightGuard:
The version of Ruby we're using is old enough that some of the gems we use will no longer install, as you can see in our build environments with tests failing.
I'm bumping the version of Ruby here to allow those gems to install.
Using the latest version of Ruby 2.3 and rubygems. This should better mirror our other environments as well.


RubyGemTSL:
@LightGuard I've merged your changes into my PR. I've also made changes to the test related Dokerfiles. They now inherit from a test base image, and then the test specific dockerfile will install only the dependancies that that particular test type requires. Let's see how my PR goes.

LightGuard:
@RubyGemTSL has added this to his PR, closing.

PR #2038, opened 2017-09-18
DEVELOPER-4435 - Scrub term before printing to page

KamiQuasi:
https://issues.jboss.org/browse/DEVELOPER-4435


KamiQuasi:
@LightGuard Jiri has approved this fix. Take a look and let me know if there is anything glaring.

PR #2047, opened 2017-09-27
DEVELOPER-3782 Updated global footer

adelasofia:


javascripts/footer.js
```diff
@@ -0,0 +1,9 @@
+jQuery(function() {
```
KamiQuasi: I'd like to see the hide/show for mobile done a little differently if possible. This is not really responsive since it would only happen on page load, so if someone had their phone in landscape mode and then rotated the content would still be hidden. Check with UX, but if we're saying they'll be expanded by default on mobile I don't see any reason to make them able to be toggled to a collapsed state.

PR #2048, opened 2017-09-28
DEVELOPER-4416 Fixes click zones on homepage promos

adelasofia:
[DEVELOPER-4416 Multiple click zones for homepage promos](https://issues.jboss.org/browse/DEVELOPER-4416)

### Review Steps
Navigate to the home page
Inspect a promo
Structure should be an anchor tag with an image and text inside (not an additional span)
Hover effect for the area should reduce the opacity of the image, but text should stay white

### DIFF
--- a/index.html.slim
+++ b/index.html.slim
@@ -11,21 +11,21 @@ description: The simple, modern & productive way to build apps and infrastructur
   .row
     .large-24
       .spotlights-container
-        a.spotlight-item href='https://developers.redhat.com/blog/?p=438176'
+        a.spotlight-item(href='https://developers.redhat.com/blog/?p=438176')
           img(src="https://static.jboss.org/images/rhd/promo/RHDP_homepromo_dotnetcore2announce_21aug2017" alt=".NET Core 2 Img")
-          span .NET Core for RHEL and the OpenShift Container Platform.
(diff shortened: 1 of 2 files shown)


RubyGemTSL:
retest this please

index.html.slim
```diff
@@ -13,19 +13,19 @@ description: The simple, modern & productive way to build apps and infrastructur
       .spotlights-container
         a.spotlight-item href='https://developers.redhat.com/blog/?p=438176'
           img(src="https://static.jboss.org/images/rhd/promo/RHDP_homepromo_dotnetcore2announce_21aug2017" alt=".NET Core 2 Img")
-          span .NET Core for RHEL and the OpenShift Container Platform.
+          p .NET Core for RHEL and the OpenShift Container Platform.
```
KamiQuasi: This should just be text instead of a block-level paragraph tag contained inside an inline anchor tag. Instead of putting the margin-bottom on the paragraph just target these classes of anchor tags. I think that will be more likely to eliminate the two-zone clicktale tracking.

adelasofia:
retest this please

adelasofia:
e2e failed tests are unrelated to this change

PR #2057, opened 2017-10-05
DEVELOPER-3782 Updated global footer

adelasofia:


RubyGemTSL:
retest this please

RubyGemTSL:
@adelasofia if you see the screenshot of the failure the footer doesn't look right.

KamiQuasi:
@adelasofia I think the test may be failing because the links are hidden on mobile. Also, you may want to double-check around the breakpoint that you've put in, because the Window width and the style's width have a point where they don't intersect (around 990px or so) and the navigation goes to the mobile layout but the toggles are unavailable.

adelasofia:
@RubyGemTSL I think the problem with the failing test is that the scenario in footer.feature is not applicable anymore (at least at the window size the test is being run at).  The links are collapsed at that size so the test can't find the link, what would be the best way to fix this? I can change the scenario to look for the toggle label instead (ex: 'COMMUNICATE').

RubyGemTSL:
@adelasofia @KamiQuasi merging this has caused all other PR's to fail due to the footer failure. I'll have to update the tests tomorrow for this.

PR #2063, opened 2017-10-10
Fix for broken footer test caused by 2057

RubyGemTSL:


adelasofia:
@RubyGemTSL thanks for getting this fix in. When Luke merged the other PR it had passed the Ruby tests, just not an e2e test related to onebox so it might have been a false positive.

RubyGemTSL:
@adelasofia np, this is what happens when the tests fail regularly, It causes confusion - these things happen.

RubyGemTSL:
well my fix didn't work @adelasofia. I will keep trying. It worked on my machine ;-) Not sure what's causing it

RubyGemTSL:
@adelasofia I added the fix for the footer in this PR instead as I'm trying to fix the search tests too. Will be fixed shortly (hopefully!!)

PR #2064, opened 2017-10-11
Flaky test fixes

RubyGemTSL:


RubyGemTSL:
@adelasofia I added the fix for the footer in this PR instead as I'm trying to fix the search tests too. Will be fixed shortly (hopefully!!)

PR #2090, opened 2017-10-30
DEVELOPER-4410 - Added promotions exclude checkbox back to the config

Dantheman720:
This had been removed in a previous commit. Added it back so that we can exclude promotion pages from DCP indexing.


Dantheman720:
retest this please

LightGuard:
Looking good, failed ruby tests, but it looks like a false positive.

Dantheman720:
@LightGuard - Yeah, the tests shouldn't be affected by this PR at all.

PR #2109, opened 2017-11-13
DEVELOPER-4081 Related product card

LightGuard:
JIRA: https://issues.jboss.org/browse/DEVELOPER-4081

Also styling it at the field level.


LightGuard:
You should be able to view it on the drupal pr site at /node/37405

PR #2116, opened 2017-11-16
Various Drupal admin config changes

LightGuard:
Setting emails back up.


LightGuard:
You should be able to make a content change, save it and look for emails attempting to go out (or maybe even going out anyway).

PR #2122, opened 2017-11-21
Removing tables from before lightning migration

LightGuard:
All the old tables (and data) are now gone. Nothing really to test here,
besides making sure Drupal is still working.


LightGuard:
Probably be good to get a rebuild of Drupal if this doesn't get merged before the nightly build.

PR #2133, opened 2017-11-29
Fix for no next session breaking the page

KamiQuasi:
Appears to be an issue where if there is not a "next" session the page breaks. Should come back in now.


RubyGemTSL:
@KamiQuasi I've fixed the SO test in #2134. Not related to your PR.

KamiQuasi:
retest this please

PR #2134, opened 2017-11-29
Fixes for flaky footer and stack overflow tests.

RubyGemTSL:


RubyGemTSL:
looks like Blinkr is going to timeout without completing. I think this should go in asap, no point retesting @LightGuard @KamiQuasi as it fixes e2e tests.

PR #2137, opened 2017-11-29
Fix coding resource display

LightGuard:
Create a coding resource, verify it displays properly (meaning it displays pretty much the same as other quickstarts). You'll notice a difference in the target product display, but that should be fine.

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/config/sync/pathauto.pattern.coding_resource.yml
+++ b/_docker/drupal/drupal-filesystem/web/config/sync/pathauto.pattern.coding_resource.yml
@@ -7,16 +7,16 @@ dependencies:
 id: coding_resource
 label: 'Coding Resource'
 type: 'canonical_entities:node'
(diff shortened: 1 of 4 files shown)


KamiQuasi:
retest this please

adelasofia:
retest this please

LightGuard:
If the broken links checks don't finish, I don't think that's a problem. @RubyGemTSL, what are your thoughts?

RubyGemTSL:
@LightGuard yes, unfortunately phantomjs uses a lot of resource and then crashes which is out of my control, it's a known issue with the phantomjs team but they no longer maintain it. I need to get some time to really dig deeper into the root cause or switch to headless chrome/firefox. I'll need to do this asap as it appears to be happening more and more recently.

RubyGemTSL:
But to answer your question - I don't think it's necessary to do a retest when the broken link checks don't finish.

PR #2145, opened 2017-12-07
DEVELOPER-4542: Fix e2e cross browser tests

RubyGemTSL:
[DEVELOPER-4542](https://issues.jboss.org/browse/DEVELOPER-4542)

### DIFF
--- a/_docker/tests/test-runner/test_run_e2e_tests_options.rb
+++ b/_docker/tests/test-runner/test_run_e2e_tests_options.rb
@@ -113,12 +113,12 @@ class TestRunTestOptions < MiniTest::Test
 
   def test_docker_execution_specifying_mobile_device
 
-    test_configuration = @run_tests_options.parse_command_line(%w(--e2e --base-url=http://foo.com --use-docker --browser=iphone_8))
+    test_configuration = @run_tests_options.parse_command_line(%w(--e2e --base-url=http://foo.com --use-docker --browser=iphone_6))
 
     assert(test_configuration[:docker])
-    assert_equal('iphone_8', test_configuration[:browser])
+    assert_equal('iphone_6', test_configuration[:browser])
     assert_equal('npm test -- --baseUrl=http://foo.com', test_configuration[:run_tests_command])
-    assert_equal('iphone_8', ENV['RHD_JS_DRIVER'])
+    assert_equal('iphone_6', ENV['RHD_JS_DRIVER'])
     assert_equal('chrome', test_configuration[:docker_node])
     assert_equal(nil, ENV['github_status_sha1'])
     assert_equal(nil, ENV['github_status_context'])
--- a/_docker/tests/test-runner/test_run_e2e_tests_options.rb
+++ b/_docker/tests/test-runner/test_run_e2e_tests_options.rb
@@ -113,12 +113,12 @@ class TestRunTestOptions < MiniTest::Test
 
   def test_docker_execution_specifying_mobile_device
 
-    test_configuration = @run_tests_options.parse_command_line(%w(--e2e --base-url=http://foo.com --use-docker --browser=iphone_6))
+    test_configuration = @run_tests_options.parse_command_line(%w(--e2e --base-url=http://foo.com --use-docker --browser=iphone_8))
 
     assert(test_configuration[:docker])
-    assert_equal('iphone_6', test_configuration[:browser])
+    assert_equal('iphone_8', test_configuration[:browser])
     assert_equal('npm test -- --baseUrl=http://foo.com', test_configuration[:run_tests_command])
-    assert_equal('iPhone 6', ENV['RHD_JS_DRIVER'])
+    assert_equal('iphone_8', ENV['RHD_JS_DRIVER'])
     assert_equal('chrome', test_configuration[:docker_node])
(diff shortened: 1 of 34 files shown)


_tests/e2e/config/chimp-local-docker.js
```diff
@@ -1,34 +1,22 @@
...
-    } catch (e) {
-        fs.mkdirSync(directory);
-    }
-}
-
 if (typeof process.env.RHD_BASE_URL !== 'undefined') {
     baseUrl = process.env.RHD_BASE_URL
 } else {
-    process.env.RHD_BASE_URL = 'https://developers.stage.redhat.com';
+    process.env.RHD_BASE_URL = 'https://developers-pr.stage.redhat.com/pr/2143/export';
```
LightGuard: This doesn't look right. Looks like a hardcoded path slipped through.

_tests/e2e/features/step_definitions/hooks.js
```diff
@@ -15,26 +15,26 @@ const hooks = function () {
...
+        //     let dirSize = [];
+        //     fs.readdirSync(pathToChromeDownloads).forEach(file => {
+        //         dirSize.push(file)
+        //     });
+        //     if (dirSize.length > 0) {
+        //         fs.emptyDir(pathToChromeDownloads)
+        //     }
+        // }
+        driver.execute('window.localStorage.clear();');
+    })
```
LightGuard: Any reason to keep this in instead of deleting it?

_tests/e2e/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM test-base:2.3.0
 
 WORKDIR /tmp
-RUN yum install -y bzip2 fontconfig freetype freetype-devel fontconfig-devel libstdc++ && \
+RUN yum install -y bzip2 fontconfig freetype freetype-devel fontconfig-devel libstdc++ unzip && \
```
LightGuard: Ideally, this would include a `&& yum clean all`

_tests/e2e/config/browsers/remote_browsers.json
```diff
@@ -1,34 +1,42 @@
 {
+  "bs_mac_safari": {
+    "browserName": "Safari",
+    "browser_version": "11.0",
+    "os": "OS X",
+    "os_version": "High Sierra",
+    "resolution": "1024x768",
```
LightGuard: If we end up using this for visual checks, we should probably include some other resolutions as well, but that's probably down the road a bit.

RubyGemTSL: Yeah don't think we need to worry about that at the moment. These are just browserstack capabilities. They can easily be reconfigured.

RubyGemTSL:
retest this please

RubyGemTSL:
@LightGuard I ended up having to do a bit more work on this after your review. it's been through and passed once, I'm just retesting to double check then it should be good to merge. Thank's.

RubyGemTSL:
All good if you're happy @LightGuard

_docker/tests/test-runner/test_run_e2e_tests_options.rb
```diff
@@ -113,12 +113,12 @@ def test_docker_execution_specifying_update_github_status
 
   def test_docker_execution_specifying_mobile_device
 
-    test_configuration = @run_tests_options.parse_command_line(%w(--e2e --base-url=http://foo.com --use-docker --browser=iphone_6))
+    test_configuration = @run_tests_options.parse_command_line(%w(--e2e --base-url=http://foo.com --use-docker --browser=iphone_8))
 
     assert(test_configuration[:docker])
-    assert_equal('iphone_6', test_configuration[:browser])
+    assert_equal('iphone_8', test_configuration[:browser])
```
LightGuard: Is the latest version of the phone (yes, I know the X is out) what we want to be testing against all the time? Seems a bit bleeding edge to me, but maybe there's a good reason

_tests/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM centos:7.3.1611
 MAINTAINER ihamilto@redhat.com
 
-RUN yum install -y  epel-release gcc gcc-c++ git libcurl-devel libffi-devel libtool libxml2 libxml2-devel libxslt \
+RUN yum clean all && yum install -y  epel-release gcc gcc-c++ git libcurl-devel libffi-devel libtool libxml2 libxml2-devel libxslt \
```
LightGuard: :) Sorry, the clean should be the last command. The layered file system of Docker takes a snapshot of the filesystem at the end of each command and that becomes the new layer. You want those layers to be as small as possible when the command finishes. In this case, we'd want to clean up any yum caches, temp files, or downloaded files at the end of the run command so we only have on that layer what is absolutely necessary. I hope that makes more sense.

_tests/e2e/Dockerfile
LightGuard: Same thing here.

LightGuard: We'd also want to remove the downloaded tarball at the end of this string of commands.

RubyGemTSL:
@Dantheman720 that failing test is not related to your PR. I finally found the root cause of that failure, and was fixed as part of #2145.

PR #2148, opened 2017-12-11
DEVELOPER-4508 - Links in footer disappear after being clicked

KamiQuasi:
https://issues.jboss.org/browse/DEVELOPER-4508

Footer links should no longer expand or collapse when the window is over 1170 pixels wide.

### DIFF
--- a/javascripts/footer.js
+++ b/javascripts/footer.js
@@ -1,6 +1,6 @@
 $(function() {
   $(".collapsible-section h3").on("click", function(){
-    var windowsize = $window.width();
+    var windowsize = document.body.clientWidth;
     if (windowsize <= 1170) {
       $(this).parent().toggleClass("collapsed");
     } else {
@@ -12,7 +12,7 @@ $(function() {
 var $window = $(window);
 
 function checkWidth() {
-    var windowsize = $window.width();
+    var windowsize = document.body.clientWidth;;
 
     if (windowsize <= 1170) {
       $(".collapsible-section").addClass("collapsed");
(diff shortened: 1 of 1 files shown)


LightGuard:
In addition to my comment, I'm still seeing links to the staging URL when I click the buttons on articles. For cheatsheets, I'm seeing some URLs pointing to the exported PR URL and some to the Drupal PR environment. Also the same for videos.

LightGuard:
This looks all kinds of messed up according to the ticket. When I'm larger than 1024 the links don't even work. When I'm below 1024 the headers collapse again once I've clicked on a link, which if I understand correctly, is not the expected behavior per the ticket.

LightGuard:
Sorry, 1170, not 1024

javascripts/footer.js
```diff
@@ -1,6 +1,11 @@
 $(function() {
   $(".collapsible-section").on("click", function(){
-    $(this).toggleClass("collapsed");
+    var windowsize = $window.width();
+    if (windowsize <= 1170) {
+      $(this).toggleClass("collapsed");
+    } else {
+      return false;
```
LightGuard: This line right here is what's stopping the default action on the link from happening.

javascripts/footer.js
```diff
@@ -1,6 +1,11 @@
 $(function() {
   $(".collapsible-section").on("click", function(){
-    $(this).toggleClass("collapsed");
+    var windowsize = $window.width();
+    if (windowsize <= 1170) {
+      $(this).toggleClass("collapsed");
```
LightGuard: Really not sure if we want this behavior or not. Do we want to collapse the section once they've clicked on a link?

KamiQuasi:
retest this please

PR #2149, opened 2017-12-11
DEVELOPER-4528 - Fix buttons to share pages on social media

KamiQuasi:
https://issues.jboss.org/browse/DEVELOPER-4528

Removed Socialite plugin, replaced with AddToAny (https://www.addtoany.com/buttons/for/website)


_docker/drupal/drupal-filesystem/web/themes/custom/rhd/templates/share.html.twig
```diff
@@ -1,24 +1,13 @@
...
-               href="http://www.linkedin.com/shareArticle?mini=true&url={{ url }}&title={{ label|render|striptags|url_encode }}"
-               data-url="{{ url('<current>') }}" data-counter="right" rel="noopener noreferrer nofollow"
-               target="_blank"><span class="vhidden">Share on LinkedIn</span></a>
-        </li>
-        <li><a class="socialite googleplus-one" href="https://plus.google.com/share?url={{ url }}" data-size="tall"
-               data-annotation="inline" data-width="60" data-href="{{ url }}" rel="noopener noreferrer nofollow" target="_blank"><span
-                        class="vhidden">Share on Google+</span></a></li>
-        </li>
-    </ul>
+<!-- AddToAny BEGIN -->
```
LightGuard: Do you want this comment to show up in the resulting HTML? If not, use {# .. #}

javascripts/social.js
```diff
@@ -1,14 +0,0 @@
-/*
```
LightGuard: Don't we need to make some changes to the gulp script to not add these files to the concatenation?

LightGuard:
In addition to my comment, I'm still seeing links to the staging URL when I click the buttons on articles. For cheatsheets, I'm seeing some URLs pointing to the exported PR URL and some to the Drupal PR environment. Also the same for videos.

Dantheman720:
@LightGuard @KamiQuasi - this seems to point to the relative environment in the export. I assume that means that it will point to 'developers.redhat.com' in prod?

LightGuard:
I would imagine so. If it's working for you @Dantheman720, I'm happy for it to go through. The first run I got was inconclusive, but the second time worked fine. I wanted a second run of the test.

KamiQuasi:
@Dantheman720 or @LightGuard feel free to merge this whenever. It appears to be working better than what is out there at the moment at the very least.

PR #2152, opened 2017-12-14
HOTFIX Update pathauto.pattern.coding_resource.yml

LightGuard:
This HAS to go in before the next Drupal restart.


RubyGemTSL:
@LightGuard failures not related to this PR. The root cause of the them was identified & fixed in #2145

PR #2156, opened 2017-12-18
DEVELOPER-4266,4267: KC and DM Tests

RubyGemTSL:
PR for both [DEVELOPER-4266](https://issues.jboss.org/browse/DEVELOPER-4266) and [DEVELOPER-4267](https://issues.jboss.org/browse/DEVELOPER-4267). The work for this is complete. However there is an issue with downloads not redirecting to hello-world page, and triggering the file download, instead returns a 404. @dhladky is investigating. 

I think this is fine to be merged. The work is done and these tickets have been hanging around for quite some time. Once David has a fix for the issue we can simply enable to download tests.


RubyGemTSL:
@robpblake would you mind casting your eye over this?

The download tests will not be ran at the moment die to an issue. When David has a fix I can simply enable them. I'd rather have them merged rather than getting another out of date stale branch.

_tests/e2e/package-lock.json
```diff
@@ -1,5717 +0,0 @@
-{
```
robpblake: @RubyGemTSL Sure we want to delete this? My understanding was that we should be comitting `package-lock.json` for wherever we're using Node in a  similar fashion to `Gemfile.lock` wherever we're using Ruby?

RubyGemTSL: 👍 pushed it back.

PR #2160, opened 2018-01-05
DEVELOPER-4521 Fix help text not showing on Drupal fields

adelasofia:
## [DEVELOPER-4521 Fix help text not showing on Drupal fields](https://issues.jboss.org/browse/DEVELOPER-4521)
There is a known bug in Drupal affecting the Help Text for link fields that have the Link Text subfield disabled. Drupal issue: https://www.drupal.org/node/2421001

### Steps to review

1. Go to create a new video resource in /node/add/video_resource
2. The SlideShare Source field should display the help text _Src attribute content from the embed code provided by SlideShare. For a more detailed explanation on where to get this link, see this doc._

PR #2164, opened 2018-01-09
Removing comment module

LightGuard:
This helps with our issue with always having updates, which there shouldn't be.
The comment module update die apparently and never finish, which is why there are always updates that need to happen.

This PR removes it so subsequent builds will not need to perform the updates. There will be an update like normal, then one without the comment updates which will probably take a bit as paragraphs has quite a bit to update, then there won't be any more updates.

You could test this by logging into the Drupal docker image and running `drush updb --entity-updates` a couple of times to verify what I said is actually happening.


LightGuard:
@Dantheman720 we need this to merge Adela's fix in #2160.

PR #2165, opened 2018-01-10
 DEVELOPER-4554 - Create Learning Path Content Type

Dantheman720:
https://issues.jboss.org/browse/DEVELOPER-4554

Review requirements document. AFAIK - We don't yet have a conclusion on points 13-17. Validation was added for long description and short description, so if you exceed the max characters for those, it should not allow you to save.


adelasofia:
@Dantheman720 Let me know when mike or tiffany approve (if they need to) this and I'll merge it

Dantheman720:
@tiffanynolan - I added a content item for #13. It can be modified in the near future if it's not exactly right, but my thought was to get this merged so that @adelasofia can begin work on the display.

PR #2167, opened 2018-01-10
DEVELOPER-4593: Fixed issue with broken link checker on PR's

RubyGemTSL:
DEVELOPER-4593: Fixed issue with broken link checker on PR's incorrectly updating the git status twice.

### DIFF
--- a/_tests/node_jenkins_test_runner.rb
+++ b/_tests/node_jenkins_test_runner.rb
@@ -23,17 +23,9 @@ class NodeJenkinsTestRunner
         tests_passed &= execute_e2e(profile)
       end
     else
-      if @host_to_test.to_s.include?('pr.stage')
-        generate_critical_link_sitemap
-        ['config/critical_links_blinkr.yaml', "#{read_env_variable('CONFIG')}"].each do |config|
-          tests_passed &= execute_blc(config)
-          break unless tests_passed
-        end
-      else
-        tests_passed &= execute_blc(read_env_variable('CONFIG'))
-      end
-      tests_passed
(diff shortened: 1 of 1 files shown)


robpblake:
@RubyGemTSL This approach won't solve the problem that we're seeing. 

The fundamental issue here is that we're executing this statement twice as independent processes:

`ruby _tests/run_tests.rb --update-github-status=...`

Each time we run the above command, on completion an update will be sent to GitHub depending on the outcome of the command. `success` if the command exited with a zero exit code, `failure` otherwise. So essentially we're performing two updates to the same GitHub status, but via different processes. The status is :

1. Updated depending on the outcome of running the above command for the "critical links"
2. Updated depending on the outcome of running blinkr on the rest of the site

If command 1 passes but then command 2 times out, then the second update is never made meaning that theoretically we get a false positive that everything is OK, when in fact I could have introduced a load of broken links, but the checking process for the rest of the site just never finished.

To resolve this we need to do the GitHub status update only after both checks have completed e.g.

```
success = false
success = check_critical_links
success = check_site_links if success
update_github_status(success)
```

In terms of how to solve this, it feels to me that the best way is to simply run Blinkr once, combining the "critical" links with the rest of the site checking. Essentially we union the existing sitemap.xml with the "critical" links sitemap.xml and then let Blinkr do it's thing. 

I'm not totally up on what the critical links are or why they need to be separated out, so let's have a chat about it when you're online.

cheers,

Rob

RubyGemTSL:
closing as will be part of #2171

PR #2170, opened 2018-01-12
[DEVELOPER-4596] Ensure Drupal shuts down gracefully on `docker stop`

robpblake:
This PR updates the `run-httpd.sh` Drupal start-up script to correctly handle the `SIGTERM` signal received from the Docker daemon when the Drupal container is being stopped.

Currently we do not handle the `SIGTERM` from the Docker daemon correctly, meaning that Docker forcefully terminates the Drupal container with a `SIGKILL` after 10 seconds (the default timeout period for a container to exit after being sent the `SIGTERM` signal). The problem with this is that Drupal might be doing some work when it gets abruptly terminated, and we believe this may have been a partial cause of the production outage we had the other day were it appears that Drupal caches became corrupt.

**To test:**

1.  Using the current `master` branch, launch Drupal in a Docker container. Once it is running, issue a `docker stop drupaldev_drupal_1` and notice it takes ~10 seconds for the `docker stop` command to return.

2. Using this PR branch, do the same and notice that the `docker stop` command returns almost immediately.

### DIFF
--- a/_docker/drupal/drupal-filesystem/scripts/run-httpd.sh
+++ b/_docker/drupal/drupal-filesystem/scripts/run-httpd.sh
@@ -1,6 +1,8 @@
 #!/bin/bash
 set -e
 
+trap "apachectl -k graceful-stop" TERM INT
+
 cd /var/www/drupal
 
 # This sleep seems to be required for certain Docker/OS combos as the bind mounts may not be fully ready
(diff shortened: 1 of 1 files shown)


robpblake:
retest this please

LightGuard:
@Dantheman720 can you run the tests Rob suggests on your box, please?

Dantheman720:
@robpblake @LightGuard - The PR version terminated immediately for me. From what I can tell, this works.

PR #2171, opened 2018-01-13
[HOLD] DEVELOPER-4603: Move critical link checks to export

RubyGemTSL:


RubyGemTSL:
@robpblake any idea what is going on here? The checks claim that all the critical pages do not exist in the sitemap:

https://redhat-dev-jenkins.rhev-ci-vms.eng.rdu2.redhat.com/job/Pull_Request_Build/3037/console

robpblake:
@RubyGemTSL The critical pages files contains the links with a trailing slash. Within the sitemap.xml provided by Drupal, there is no trailing slash hence the failure i.e. it is comparing `/promotions/the-null` to `/promotions/the-null/` and saying it doesn't exist.

We should probably update the code to cope with this as I suspect that will be a source of annoyance once we're gone.

_docker/lib/export/export_inspector.rb
```diff
@@ -45,11 +51,37 @@ def check_if_path_exists(url, expected_file_path, export_directory)
     expected_path = File.join export_directory, expected_file_path
     exists = File.exist?(expected_path)
     unless exists
-      puts "WARNING: Cannot locate export of '#{url}' at expected path: '#{expected_path}'"
+      puts"WARNING: Cannot locate export of '#{url}' at expected path: '#{expected_path}'"
```
LightGuard: Uh... whitespace?

_docker/lib/export/export_inspector.rb
```diff
@@ -78,7 +110,7 @@ def inspect_export(url_list, export_directory)
       puts "WARNING: Of '#{total_pages}' pages, '#{missing_pages}' are not found in the export."
 
       if should_fail_on_missing_content
-        raise StandardError.new("There are '#{missing_pages}' pages missing from the site export (see warnings for missing content). Failing export process.")
+        raise(StandardError,"There are '#{missing_pages}' pages missing from the site export (see warnings for missing content). Failing export process.")
```
LightGuard: I think for consistency we should use the one without parens (pretty sure that's what is being used in other places)

RubyGemTSL:
retest this please

RubyGemTSL:
@robpblake if you're happy can we get this merged?

robpblake:
@RubyGemTSL It looks like we're still doing critical link checking in Blinkr via a sanity-check profile?

RubyGemTSL:
@robpblake yes, I'd like to keep this as and at least have the ability to run these sanity checks in production. We can agree when we run these checks at a later date?

PR #2173, opened 2018-01-17
[DEVELOPER-4597] [DEVELOPER-4602] Refactored Drupal start-up script 

robpblake:
This PR refactors the `DrupalInstallChecker` class to remove un-required code.  It updates the code to make the assumption that Drupal has already been installed and that we are applying the latest configuration from Git to the current installation.

Additionally  we now perform a `drush cr` before attempting any other Drush commands to import the current Drupal configuration (using this approach solved our production outage last week).

The boot sequence is now:

1. Wait until the database is available for connections
2. Verify that `rhd.settings.php` and `settings.php` exist in the expected Drupal filesystem location
3. Verify that expected tables exist within the Drupal database
4. Perform a `drush cr`
5. Perform a `drush cim`
6. Perform a `drush cr`
7. Perform a `drush updb`
8. Perform a `drush cr`

Finally, this PR also resolves DEVELOPER-4602 by providing a clear error message if the boot of Drupal is aborted because of missing database tables.

**To test:**

- Ensure that Drupal boots as usual
- To witness a failure, you could for example drop one of the required tables from the MySQL instance Drupal is booting against e.g. `drop table lightning_node`

### DIFF
--- a/_docker/drupal/drupal-filesystem/scripts/drupal_install_checker.rb
+++ b/_docker/drupal/drupal-filesystem/scripts/drupal_install_checker.rb
@@ -21,148 +21,107 @@ class DrupalInstallChecker
     @opts = opts
   end
 
-  def settings_exists?
-    File.exists? File.join drupal_site, 'settings.php'
+  def prepare_drupal_for_boot
+
+    wait_for_database_to_boot
+    check_all_required_drupal_configuration!
+
+    drush_clear_cache
+    drush_import_config
+    drush_update_db
   end
 
-  def rhd_settings_exists?
-    File.exists? File.join drupal_site, 'rhd.settings.php'
+  private
+
+  def ensure_file_exists!(file_name)
+    expected_file = File.join(@drupal_site, file_name)
+    puts "Checking for presence of required Drupal configuration file '#{expected_file}'..."
+
+    raise StandardError "Required Drupal configuration file '#{expected_file}' is missing. Aborting Drupal boot process." unless File.exists?(expected_file)
+
+    puts "\tFound Drupal configuration file '#{expected_file}.'"
   end
 
-  def mysql_connect?
+  def settings_exists!
+    ensure_file_exists!('settings.php')
+  end
+
+  def rhd_settings_exists!
+    ensure_file_exists!('rhd.settings.php')
+  end
+
+  def wait_for_database_to_boot
+    sleep (5) until database_connect?
+  end
+
+  def database_connect?
+
+    puts "Checking connectivity to database '#{@opts['database']['host']}:#{@opts['database']['port']}'..."
+
     begin
       process_executor.exec! 'mysql', ["--host=#{@opts['database']['host']}",
-                                       "--port=#{@opts['database']['port']}",
-                                       "--user=#{@opts['database']['username']}",
-                                       "--password=#{@opts['database']['password']}",
-                                       '--connect-timeout=20',
-                                       "#{@opts['database']['name']}"]
+      "--port=#{@opts['database']['port']}",
+      "--user=#{@opts['database']['username']}",
+      "--password=#{@opts['database']['password']}",
+      '--connect-timeout=20',
+      "#{@opts['database']['name']}"]
+
+      puts "\tSuccessful connection to database '#{@opts['database']['host']}:#{@opts['database']['port']}'."
       true
     rescue => e
-      puts "ERROR: #{e.message}"
+      puts "Failed to make connection to database '#{@opts['database']['host']}:#{@opts['database']['port']}', reason: #{e.message}"
       false
     end
   end
 
-  def tables_exists?
-    begin
-      tables_to_check = %w(lightning_node lightning_node__body lightning_taxonomy_index)
-      tables = process_executor.exec!('mysql', ["--host=#{@opts['database']['host']}",
-                                                "--port=#{@opts['database']['port']}",
-                                                "--user=#{@opts['database']['username']}",
-                                                "--password=#{@opts['database']['password']}",
-                                                '--execute=show tables', "#{@opts['database']['name']}"]).split("\n")[1..-1]
-      return false if tables.nil? || tables.empty?
-      (tables_to_check.uniq.sort - tables.uniq.sort).empty?
-    rescue => e
-      puts "ERROR: #{e.message}"
-      false
+  def database_tables_exists!
+    required_tables = %w(lightning_node lightning_node__body lightning_taxonomy_index)
+    puts "Checking for required database tables '#{required_tables}' in Drupal database..."
+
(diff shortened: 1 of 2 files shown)


robpblake:
@LightGuard I think you're best placed to review this one.

cheers,

Rob

_docker/drupal/drupal-filesystem/scripts/drupal_install_checker.rb
```diff
@@ -21,148 +21,107 @@ def initialize(drupal_site, process_executor = ProcessExecutor.new, opts)
...
-                                       '--connect-timeout=20',
-                                       "#{@opts['database']['name']}"]
-      true
-    rescue => e
-      puts "ERROR: #{e.message}"
-      false
-    end
+  private
+
+  def ensure_file_exists!(file_name)
```
LightGuard: Why are you using the "!" at the end? we're not really changing the internals of the object, and it doesn't return itself.

_docker/drupal/drupal-filesystem/scripts/drupal_install_checker.rb
```diff
@@ -21,148 +21,107 @@ def initialize(drupal_site, process_executor = ProcessExecutor.new, opts)
...
-                            '--config-dir=/var/www/drupal/web/config/sync',
-                            '-y'])
+  def drush_clear_cache
+    process_executor.exec!('/var/www/drupal/vendor/bin/drush', %w(--root=/var/www/drupal/web cr))
   end
 
-  def update_db
-    puts 'Executing drush dbup'
+  def drush_update_db
+    puts 'Executing drush dbup...'
```
LightGuard: this should actually be updb

LightGuard: we could even drop it and manually do an updb when we know we need to.

_docker/drupal/drupal-filesystem/scripts/drupal_install_checker.rb
robpblake: @LightGuard My understanding is that the exclamation mark at the end of the method name was used to denote methods in Ruby that were considered "dangerous" e.g. modified the receiver or raised an Exception. 

I've gone down the route of using the `!` if a method raises an Exception. Happy to remove this if it's not really idiomatic of Ruby as I'm still approaching a lot of this with my Java hat on :-)

_docker/drupal/drupal-filesystem/scripts/drupal_install_checker.rb
robpblake: @LightGuard I will re-name to reflect that this is actually `updb` as you say.

In terms of removing the `updb` call entirely from the boot sequence of Drupal, can you remind me what was the motivation behind doing that?

We would need to be aware that doing that will cause us a headache with the PR pipeline in that people would have to make sure that they ran the `updb` on their PR Drupal instance each time their PR was built if they are pushing changes to Drupal or the modules we use. I suspect people would forget to do that, particularly in the case of a "retest this please".

cheers,

Rob

_docker/drupal/drupal-filesystem/scripts/drupal_install_checker.rb
LightGuard: Here's what I found:

Bang (!) Methods

Ruby methods that modify an object in-place and end in an exclamation mark are known as bang methods. By convention, the bang labels a method as dangerous - specifically, as the dangerous equivalent of a method with the same name but without the bang.

You'll find a number of pairs of methods, one with the bang and one without. Those without the bang perform an action and return a freshly minted object, reflecting the results of the action (capitalizing a string, sorting an array, and so on). The bang versions of the same methods perform the action, but they do so in place: Instead of creating a new object, they transform the original object.

Examples of such pairs of methods include sort/sort! for arrays, upcase/upcase! for strings, chomp/chomp! for strings, and reverse/reverse! for strings and arrays. In each case, if you call the non-bang version of the method on the object, you get a new object. If you call the bang version, you operate in-place on the same object to which you sent the message.

In Ruby you can define a method name that ends with an exclamation point or bang. The bang methods are called and executed just like any other method. However, by convention, a method with an exclamation point or bang is considered dangerous.

Normally for the built-in classes, dangerous usually (although not always) means this method, unlike its non-bang equivalent, permanently modifies its receiver.

You'll find a number of methods, one with the bang and one without. Those without the bang perform an action and return a new object. The bang versions of the same methods perform the action, but they do so in place: Instead of creating a new object, they transform the original object.

A few non-bang methods perform changes on the original string. The names of these methods make it clear that this is happening (such as replace), even though there's no ! on the name.

LightGuard: If you feel it falls in that definition, okay, cool, I can go with that.

robpblake: @LightGuard Updated to remove the `!` from methods as I don't see them being "dangerous" as per that description above.

cheers,

Rob

_docker/drupal/drupal-filesystem/scripts/drupal_install_checker.rb
robpblake: @LightGuard OK I understand.

At the minute I don't really see us running the `updb` as a problem as it only takes ~10 seconds when there are no changes to apply. As long as it's not breaking anything (which I assume it isn't as we've been doing it for a while), then my preference would be to leave it as is simply because it means that at the minute nobody needs to remember to apply any changes. 

I think the biggest challenge of making it a manual action would be ensuring people have access and remember to apply it on their PR builds.  We would ideally want people to apply any module modifications before the export starts so that we can ensure a module update doesn't break parts of the site. We'd therefore be looking at some sort of manual step in the PR pipeline that says "Have you run updb if you need to?" . The build would be paused until it is confirmed `updb` has been applied. It's possible to do that with Jenkins pipeline builds, but it doesn't feel like a great addition to the PR workflow.

Perhaps another alternative is that we somehow provide a flag to indicate to the Drupal boot process that it should apply `updb`  this time around, but again this feels like something to consider when running `updb` is causing a problem?

cheers,

Rob

LightGuard: Okay, that's fair. We can keep it in.

LightGuard:
I don't know if we can get docker to spit out the stout so jenkins can log it. It would be quite helpful.

robpblake:
@LightGuard Should be simple enough I believe. We can simply update the PR Jenkins job configuration with a `docker logs` and the name of the Drupal container and that should dump the boot log of Drupal to the stdout and be captured as part of the Jenkins job.

Are you happy for this PR to be merged?

cheers,

Rob

PR #2174, opened 2018-01-19
[Blinkr] Updated Blinkr ignores

RubyGemTSL:
Simply added url's to be ignored when generating blinkr report for prod checks.


RubyGemTSL:
@robpblake this is ready to be merged. The broken links are fixed (ignored) in #2171

PR #2175, opened 2018-01-25
DEVELOPER-4609 - .NET Core download page incorrect

KamiQuasi:
https://issues.jboss.org/browse/DEVELOPER-4609

Adds additional logic to test the download manager service response to make sure an empty table isn't created.

### DIFF
--- a/javascripts/downloads.js
+++ b/javascripts/downloads.js
@@ -227,7 +227,7 @@ $(function() {
 
   if($productDownloads && productCode) {
     $.getJSON(app.downloads.url + productCode,function(data) {
-      if(!data.length) {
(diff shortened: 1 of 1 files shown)


KamiQuasi:
One way to test this is to do a search in dev tools for "$.getJSON(app.downloads.url + productCode,function(data) {" and then put a break point on the first if statement following that line and type this in the console to change the data object when the breakpoint is hit:
`data = [{"name":"dotNET on RHEL","productCode":"dotnet","featuredArtifact":null,"productVersions":[]}]`

Dantheman720:
@KamiQuasi - I think we should probably remove the all downloads heading as well here when downloads don't exist.

dotnet_downloads

PR #2190, opened 2018-02-07
Fixes footer on error pages

adelasofia:


LightGuard:
@adelasofia looks like we have a bunch of whitespace between the footer and the end of our content on the homepage. Is that something you can fix here as well?

adelasofia:
@LightGuard ha! you read my mind. I just put a PR to fix that #2191

PR #2197, opened 2018-02-09
DEVELOPER-4678 - Reference keycloak JS from Keycloak server

KamiQuasi:
https://issues.jboss.org/browse/DEVELOPER-4678

Drupal library changes to reduce some clutter in the html.html.twig file and also load the keycloak.js file straight from the auth server rather than build it in our scripts.


LightGuard:
Good with these changes, @KamiQuasi any reason NOT to merge?

lkrzyzanek:
Checking the https://developers-pr.stage.redhat.com/pr/2197/export/ (not sure if it's right place) but the HTML source have following:
<script src="auth/js/keycloak/">
It needs to be /auth/js/keycloak.js. Then it would leads to https://developers-pr.stage.redhat.com/auth/js/keycloak.js

LightGuard:
@KamiQuasi  can we push this to the staging branch to try it there?

lkrzyzanek:
Checking https://developers.stage.redhat.com and there is wrong link to the JS. There is ``<script src="auth/js/keycloak/"></script>``

KamiQuasi:
@lkrzyzanek I'm checking with others to see if that know why the export would be stripping off the file extension. If we can prevent that in this case I think the change will work.

lkrzyzanek:
Ok, thanks. btw. the right URL is: <script src="/auth/js/keycloak.js"></script> (the slash at the begin is needed as well otherwise it'll work only on homepage.)

PR #2201, opened 2018-02-13
[STAGING] DEVELOPER-4678 - Remove Keycloak JS and use server's file

KamiQuasi:


paulrobinson:
@cben it would be good to understand what you are trying to achieve. My initial impression is that there's probably a better place for you to import keycloak.js from.

@KamiQuasi would you mind helping out @cben here (or ask someone else to)?

PR #2207, opened 2018-02-15
[DO NOT MERGE] First attempt at fixing keycloak

LightGuard:


LightGuard:
retest this please

LightGuard:
@KamiQuasi Please take a look.

RubyGemTSL:
let me know if you need any help with this, but the tests are failing because the login/register links aren't working. (probably what you're trying to fix here)

KamiQuasi:
@RubyGemTSL and @LightGuard this "worked" as far as the including of the path to the script file, but the functionality of the script is pointing to production without the same querystring variables, so the iframe inclusion fails to work. Hoping @lkrzyzanek might be able to sort out why.

PR #2214, opened 2018-02-20
Fixes styles for middleware page

adelasofia:
Fixed issue where the page's path wasn't being added to the body tag of all pages. This will fix any issues where we need the path as a CSS class, such as the /middleware page to have page-specific CSS.


KamiQuasi:
retest this please

PR #2221, opened 2018-02-23
Set relative paths for testing

KamiQuasi:
Product downloads should be relative now.


LightGuard:
I'm really not worried about that test that failed, @KamiQuasi are you?

PR #2227, opened 2018-02-28
DEVELOPER-4706 SPIKE for assembly DevNation Live

LightGuard:
Mostly used as a playground to try things out.


LightGuard:
@KamiQuasi If you're good with this, go ahead and merge.

PR #2234, opened 2018-03-06
Fix Product downloads

KamiQuasi:
Puts back the code for product download buttons that was removed, but keeps the duplicate service call from being made so we don't mess up the downloads page.


KamiQuasi:
retest this please

PR #2238, opened 2018-03-07
Fix for failing product tests

RubyGemTSL:
Fix for failing product tests.


RubyGemTSL:
@LightGuard this is the fix for the drupal `issues` we had yesterday. ;-)

PR #2247, opened 2018-03-14
DESIGN-1510 New homepage promo

adelasofia:
[DESIGN-1510 New homepage promo](https://issues.jboss.org/browse/DESIGN-1510)

## NOTE:
This needs to go live at the same time as the change in text/link underneath the image (a content change in Drupal)


LightGuard:
When is that change supposed to happen?

adelasofia:
@LightGuard it's not urgent but asap. If you can merge it I'll make the content change and force the drupal rebuild later on today when no one is using it.

PR #2248, opened 2018-03-15
Assembly that displays Wordpress posts with category filter

fearlsgroove:
Adds a new assembly type "Blog posts". Optionally choose one or more categories from wordpress to filter the results.

The output is currently unthemed, this basically a technology demo, but the hooks for theming are in place and a design could easily be applied in a follow up.


LightGuard:
this is ok to test

PR #2260, opened 2018-03-23
JIRA 4796 - Wordpress post entity reference

delzhand:
- Refactors existing wordpress API calls into a service.
- Adds several methods to that service to facilitate display/querying of wordpress content
- Adds a field type, widget, and formatter for a single wordpress post


delzhand:
As part of this PR, we've also created a wordpress endpoint. Once that endpoint is in place, there's a change to be made here that will give more focused results on the reference widget. This PR should remain open until that's resolved.

LightGuard:
this is ok to test

LightGuard:
How are we best able to verify this is working?

delzhand:
You can add a field of type Wordpress Post to any entity type, and then create or edit an instance of that type. Example:

1. Add a field to the Landing Page content type.
2. Create a landing page and populate the field.
3. Confirm that the selected wordpress post is displayed when viewing your landing page.

PR #2262, opened 2018-03-26
use new title search endpoint for wordpress post matching

delzhand:
This uses the new title search wordpress endpoint to improve functionality of the wordpress post autocomplete from JIRA 4796. Also fixes a bug which would prevent content without featured media from rendering properly.


LightGuard:
this is ok to test

LightGuard:
Is this ready to go?

PR #2274, opened 2018-04-04
Fix for intermittent failures on PR's.

RubyGemTSL:
Minor changes to fix intermittent failures on PR's. 

- Upped the wait time for login/register pages.
- Added try/catch to after hooks. Noticed on @KamiQuasi 's PR the after hook timed out with `function timed out after 60000 milliseconds`. Catching the after hook with console output as we don't want the test to fail after it is already complete.


RubyGemTSL:
I'm on holiday but will keep an eye on any more issues.

PR #2275, opened 2018-04-04
add fences module

delzhand:
This is a dependency for several entity display modes. Composer automatically updated an unrelated repo URL, but removing it would invalidate the hash in the lock file.


paulrobinson:
This is ok to test

PR #2277, opened 2018-04-04
DEVELOPER-4807: Featured Evangelists assembly

delzhand:
Provides Featured Evangelists assembly type. Depends on DEVELOPER-4806.


LightGuard:
this is ok to test

delzhand:
Marking this as do not merge for a moment - I need to get a commit in that prevents Fences from doing anything on the current theme.

PR #2278, opened 2018-04-04
fences shouldn't do anything when rhd is the active theme

delzhand:
Fences needs to be enabled for a lot of the frontend work we're doing with the new theme, but we don't want it to have any effect on the current them. Fences works by altering the theme registry and inserting its own path the field templates - this simply forces the path to be that of the Classy theme (what RHD inherits from)


paulrobinson:
this is ok to test

PR #2279, opened 2018-04-04
[URGENT] EventsValidationFix - Fixed date comparison bug

Dantheman720:
To test: Create a new event with dates included. This should no longer bug out.

Also, make sure that form error is thrown when end date is before start date.


LightGuard:
All good on the manual testing side.

PR #2285, opened 2018-04-05
DEVELOPER-4805: new Topic Page node type

delzhand:
There might need to be a followup issue on this - I can't add assembly types to the sections field that are currently pending pull requests. Any future tasks to create an assembly type should consider it part of that task to add the assembly type to the sections field for Landing Pages and Topic Pages.


paulrobinson:
this is ok to test

PR #2290, opened 2018-04-06
[URGENT] Uncompress rhd-frontend css

delzhand:
Because minified CSS is impossible to resolve merge conflicts on, this commit replaces rhd.min.css with rhd.css, which is now built without the "compressed" output style by node-sass.


paulrobinson:
this is ok to test

PR #2298, opened 2018-04-09
Use title field instead of name on assemblies

fearlsgroove:
We were in correctly using the assembly "Name" field as a visitor-facing bit of content.


KamiQuasi:
this is okay to test

KamiQuasi:
retest this please

PR #2309, opened 2018-04-12
DEVELOPER-4856 - Initial Navigation Menus for header and footer

KamiQuasi:
Creates the menus for navigation in the header and footer


LightGuard:
Is this ready to go?

PR #2321, opened 2018-04-16
Adding all the other twig files from the old theme

LightGuard:


LightGuard:
I think this should help us get closer to looking correct on those older pages. Something we could do with the older twig files at the node level is to do an attach a new library (the old CSS) so we only have the older CSS files loading on the older pages.

PR #2326, opened 2018-04-19
DEVELOPER-4915 Fix title display in RHDP theme

delzhand:
The title is now displayed. If the page alias is "harmonize/home", the title will be omitted.


LightGuard:
this is ok to test

PR #2350, opened 2018-04-24
Featured resources style tweaks

delzhand:
At some point _card.scss was moved from src/styles to src/styles/patterns, but the original was left in place.


LightGuard:
this is ok to test

PR #2365, opened 2018-04-25
DEVELOPER-4945/4949 dynamic content list improvements

delzhand:
Formatting for teasers as well as hiding the Latest comments section and making better use of space when nothing is in the sidebar


LightGuard:
this is ok to test

PR #2366, opened 2018-04-25
Fixes SA-CORE-2018-004

LightGuard:
New exploit was found, this fixes it.


LightGuard:
I don't see how this change affected the e2e tests

PR #2378, opened 2018-04-26
Rollbacks should happen in the export context

LightGuard:
In testing the export, I had to make the following changes for things to
actually work.


LightGuard:
Errors in blinkr are coming from /harmonize/styleguide which shouldn't even be in the sitemap.

PR #2390, opened 2018-04-30
Update image hero to keep background in container

fearlsgroove:
Moves the image into the container element and tweaks the width of the content area.

After merging, needs some content changes:

* Replace the image with a version without whitespace on either side.
* Remove line breaks


paulrobinson:
this is ok to test

PR #2432, opened 2018-05-03
Moving title up higher for monitoring checks

LightGuard:
Our monitoring solutions need to verify the correct page is loaded. They
would like something towards the top that is guaranteed to be there. The
title seemed to be a good option, but it needs to be within the first 5k
of the response. I'm moving it up higher so it's guaranteed to be in
that amount.


paulrobinson:
The title should be in this format:

Red Hat Developer | <Page Title>

LightGuard:
We changed it??

paulrobinson:
I think that's always been the plan. Not sure it got scheduled in JIRA tho.

paulrobinson:
Notice the name of the site in the logo: https://developers.stage.redhat.com/

PR #2441, opened 2018-05-03
Styles for quickstarts

adelasofia:


paulrobinson:
Looks like this PR may have broke the export. Can you test a PR that reverts this? It doesn't seem to revert automatically.

PR #2442, opened 2018-05-03
Clear the cache before an export

LightGuard:
Attempting to clear the cache before an export


LightGuard:
The last one worked, you can see in the logs that `drush cr` is being called before the export. The export died because of data issues from prod data.

PR #2470, opened 2018-05-14
Attempting to remove awestruct from the pipeline

LightGuard:


LightGuard:
This is ready now. You can go into the logs and look for awestruct being called, that's the test. For the other environments, we simply need to disable the awestruct jobs.

RubyGemTSL:
@LightGuard FYI, the tests are failing for mobile, and will continue to do so until the search filter bug is fixed.

PR #2472, opened 2018-05-15
DEVELOPER-5058 && DEVELOPER-4763 Removing awestruct take two

LightGuard:


KamiQuasi:
Think we'll need to point the tests at the new JS file to make them pass.

LightGuard:
@KamiQuasi what needs to happen? Is it an include or something somewhere?

KamiQuasi:
`/Users/ldary/rh/developers/_tests/unit/config/karma.conf.js` just needs to point from `../../../javascripts/build.js` to the rhd.min.js file in the new theme I believe.
You could also bump the version on the web components stuff up to 1.2.0 while you're in there too.

LightGuard:
@KamiQuasi any other ideas?

LightGuard:
retest this please

LightGuard:
It was being kept around as a base to look at for removing awestruct, but we can close it.

PR #2480, opened 2018-05-17
DEVELOPER-5055 - Fix filters on search mobile view

KamiQuasi:
https://issues.jboss.org/browse/DEVELOPER-5055


RubyGemTSL:
Looks like it's still not clickable @KamiQuasi, looks like filters may be open (but hidden).

RubyGemTSL:
Looks good now @KamiQuasi 👍

PR #2492, opened 2018-05-29
DEVELOPER-5096: [e2e] Infinite loop exists whilst waiting for download to exist

RubyGemTSL:
There was an infinite loop whilst waiting for the download to exist. Occasionally when downloading pdf's they do not trigger, therefore triggering an infinite loop causing the build to timeout with failure after 60 seconds.


RubyGemTSL:
retest this please

PR #2488, opened 2018-05-24
[DO NOT MERGE] DEVELOPER-5063 - Desktop and mobile menu states for logged in users

Dantheman720:

PR #2500, opened 2018-06-05
DEVELOPER-5063 Desktop and mobile menu states for logged in users

adelasofia:
[DEVELOPER-5063 Desktop and mobile menu states for logged in users](https://issues.jboss.org/browse/DEVELOPER-5063)

-- This PR replaces #2488


LightGuard:
This is looking good to me but would like another pair of eyes to verify my testing.

lkrzyzanek:
Works for me but only one UX hint.
Only name is clickable. The arrow does nothing. I would expect that whole "greyed" area when mouse hover it would be clickable not only the name.

PR #2503, opened 2018-06-14
DEVELOPER-5130 Change link in ts file from video to videos

alistairmc:
https://issues.jboss.org/browse/DEVELOPER-5130

Visit /devnationlive and click on the video links, links should work


KamiQuasi:
this is ok to test

PR #2505, opened 2018-06-19
DEVELOPER-5133 -  Change bpmsuite to rhpam in hard-coded cases

KamiQuasi:
https://issues.jboss.org/browse/DEVELOPER-5133

@paulrobinson to verify and merge in the morning
Should fix downloads and search in particular where the product code is not dynamically pulled in


paulrobinson:
@KamiQuasi I merged this as it did what it needed to do. Although I was surprised at how much code needed changing for this. Can you take a look at the files changed and see if anything needs reverting?

PR #2506, opened 2018-06-19
DEVELOPER-5124

alistairmc:
Updated the red underline on the nav to 3px, @tiffanynolan please review.


paulrobinson:
this is ok to test

PR #2510, opened 2018-06-27
Developer-5095 Add fixed height for a tag in rhd nav

alistairmc:
When vieweing this page https://developers.redhat.com/events/ you can see the correct format in Chrome and the incorrect format in Firefox.


KamiQuasi:
this is ok to test

PR #2511, opened 2018-06-27
DEVELOPER-4975 change cdn( to url(

alistairmc:
Devnation 2015 page doesn't load content and header is missing See: http://rhdp-drupal.stage.redhat.com/devnation2015

Now should have header and data


KamiQuasi:
this is ok to test

PR #2516, opened 2018-07-03
DEVELOPER-4665 - Default active tab for installation instructions should follow user's OS

Dantheman720:


LightGuard:
I cannot test this as I don't have a mac :)

LightGuard:
@adelasofia can you verify this?

PR #2519, opened 2018-07-05
[FIX] - Remove Drupal server dependency from referrer script

KamiQuasi:
Add ?referrer=jbd to pages like /middleware /products /


LightGuard:
How do I test this? Going to the /middleware page and adding ```?referrer=jbd``` resulted in a JS error stating that ```category``` is not defined

RubyGemTSL:
sorry this damn cheatsheet download test is flaky. I'm looking into this today.

PR #2520, opened 2018-07-06
DEVELOPER-5177 Add image.style.media.asset.cover_medium.yam to site config for new m…

alistairmc:
…edia image style

### JIRA Issue
https://issues.jboss.org/browse/DEVELOPER-5177

### Verification Process
The style is called "Media Cover Image", and needs to be preserved in the configuration so it can be used going forward.

Image style is maintained even after Drupal server rebuilds

Add an image from the media library and see that the new image style is available


LightGuard:
this is ok to test

PR #2526, opened 2018-07-12
Feature/developer 4891

alistairmc:
### JIRA Issue
developer 4891

### Verification Process
check site and that fonts look ok

Can this be built to test I need to test the export site and Drupal, works locally but needs testing since I have changed a lot of files @KamiQuasi can you have a look and see if this is the right approach.

I have linked to the google web fonts css and also updated the font references we seemed to have a mix of 'overpass' and 'Overpass' I have added Arial to the font fallback as this should work better on windows os.


alistairmc:
@LightGuard Is there any way to get the details of what test are failing within js-e2e-teste when I click the details I get oops! with no css and  java.lang.NullPointerException

LightGuard:
@alistairmc You're probably not on the list to be able to access the build server. It's a failed cheat-sheet download, looking into it. I'm not entirely sure what's going on with that test, but I doubt it's related. @RubyGemTSL any thoughts?

RubyGemTSL:
@LightGuard @alistairmc yes, if you see failures around cheatsheets it's related to the test user, I believe akamai is somehow blocking the download from triggering. I'm working on fixing this now. Should be finished today.

LightGuard:
Maybe it's just me, but why are we defining the same fonts in so many places?

alistairmc:
@LightGuard  I completely agree base font should be declared once for page and then cascade. Maybe a different story for that one though as not sure on the impact on web components id they are used outside developers.redhat.com

KamiQuasi:
@alistairmc and @LightGuard a lot of these will go away as we convert to using CSS variables for theming. Most of them are within the Web Components because the styles are encapsulated and without the proper specificity the font family can sometimes have unintended consequences.

PR #2529, opened 2018-07-16
DEVELOPER-5103 drop query string from url

LightGuard:
### JIRA Issue Link
* https://issues.jboss.org/browse/DEVELOPER-5103

### Verification Process
Look at the export, make sure the og:image doesn't include a reference which has a query string in it.


LightGuard:
retest this please

PR #2574, opened 2018-08-27
[DEVELOPER 5264] Media Query Cleanup (revisit)

rogeralloymagnetic:
### JIRA Issue Link
* https://issues.jboss.org/browse/DEVELOPER-5264

### Verification Process
1. Browse through all assembly types and page types and there should be no difference in styles from master branch.
2. Moving forward: all media query breakpoints and mixins will be defined in `_mq-mixins.scss` partial


LightGuard:
this is ok to test

Jason Porter
Sent from Outlook<https://aka.ms/qtex0l> on iOS
________________________________

PR #2576, opened 2018-08-28
[DEVELOPER 5309] Add audience_selection to all assemblies

ghost:
### JIRA Issue Link
https://issues.jboss.org/browse/DEVELOPER-5309

### Verification Process
1. Add or edit any assembly
2. Choose an option from "Audience Selection"
3. Save and view that that assembly is now hidden (css rule currently hides all assembly items with the "data-audience" attribute)
4. View source/inspect to see there is a new "data-audience" attribute that outputs the values of the field_audience_selection field as a comma separated list.

### TO DO for @KamiQuasi 
1. Wire up Keycloak
2. Show assemblies based on access rights

**Note:** New audience selection roles/types can be set in the future by updating the settings on the "field_audience_selection" field. This will need to be config exported and committed to git, it is not a content manager task.

### Screenshot
screen shot 2018-08-28 at 8 54 09 am


robpblake:
this is ok to test

PR #2577, opened 2018-08-29
[DEVELOPER-4890] move images to static and symlink

alistairmc:
### JIRA Issue Link
* Link to the JIRA ticket(s) this pull request solves
https://issues.jboss.org/browse/DEVELOPER-5190

### Verification Process

links on site to images should still work, no broken links


alistairmc:
retest this please

alistairmc:
retest this please

alistairmc:
@KamiQuasi @LightGuard @adelasofia Could one of you have a look at this PR it's the best I can get, Drupal site can see static images, Exported live site will be able to see images but PR export site cannot due to the /pr/****/ url this would need to be rewritten or done with a :port 

Any suggestions or do you think we could go with this since it works in the most part and is an improvement to what we have.

robpblake:
@alistairmc @KamiQuasi Can I get a bit of background on why we need to re-locate the images? The static directory in the export folder was always meant for stuff that isn't or can't be downloaded from Drupal by HTTrack (the tool that does the page crawling). With this modification what we're ending up with is HTTrack downloading all of the images and then us over-writing them with the static copy. 

It also seems a shame that we're going to break a lot of images in the PR environments with this approach.

cheers,

Rob

alistairmc:
I welcome a better approach, this one is far from ideal.

Only the images that were hosted on static.jboss should be in this static folder. Drupal continues to host most of the images this was only for content referenced in the css from the old sites.

Regards

alistairmc:
Sorry forgot to send background too, the ticket was to move all static resources off static.jboss.org <http://static.jboss.org/> so we had no cross domain issues.

Alistair McGranaghan
Online Design and Development

T: 0191 3845705
M: 07983 358051
alistair@alt-media.co.uk

robpblake:
@alistairmc Thanks! Just read through https://issues.jboss.org/browse/DEVELOPER-4890 which I assumed was the main reason for the image relocation.

As for a method of approaching this. I think the following would work:

1. Generate a list of resources to be pulled from static.jboss.org
2. Put together a small shell-script to curl/wget those resources into the existing images folder within the Git repo. This images folder is already served by the Apache instance in which we host Drupal as a static resource and we "know it works"™
3. Update the Twig layouts/CSS/anything else to refer to "/images/my-image.png" instead of "https://static.jboss.org/images/my-image.png". This will always render correctly in the context of Drupal as `/images` is served from the root of the Drupal context.
4. As the images are now being correctly served by Drupal, the export process will determine that it needs to download them (as it does with other images already managed within Drupal). The PR and other export processes should therefore continue to work as-is.
5. profit...

It means we don't need any changes to any of the Docker configs etc and it also means we don't have the duplicate image copying process as part of the export.

cheers,

Rob

alistairmc:
Is that not what I have done?

alistairmc:
retest this please

alistairmc:
@robpblake @KamiQuasi I have made changes based on Robs feedback could someone take a look and merge if possible, have reverted to /images dir at the repo root removing the need to change any mount points or symlinks, still not working for PR environment but we could not come up with a way to resolve this without access to the PR infrastructure.

alistairmc:
Sorry hold on that just pushed a merge

alistairmc:
All tests pass build seems ok

alistairmc:
@KamiQuasi @robpblake  please review and merge, thanks

PR #2605, opened 2018-09-13
Updating to latest Drupal and Lightning

LightGuard:
### JIRA Issue Link
* N/A -- Maybe I'll create one

### Verification Process
Make sure the site still works, videos stil work, adding a video, views for the DCP, adding assemblies, etc.


robpblake:
@LightGuard Looks like there is a merge conflict. Would you mind resolving please?

cheers,

Rob

_docker/drupal/drupal-filesystem/composer.json
```diff
@@ -24,11 +24,11 @@
     "require": {
         "drupal-composer/drupal-scaffold": "^2.0.0",
         "cweagans/composer-patches": "^1.6.0",
-        "acquia/lightning": "3.1.7",
+        "acquia/lightning": "3.2.0",
         "drupal/admin_toolbar": "1.24",
         "drupal/assembly": "1.x-dev",
         "drupal/asciidoc": "1.2",
-        "drupal/core": "^8.5",
+        "drupal/core": "^8.6",
```
robpblake: I personally think we should lock this down to "8.6.1" (in the same way that we're locking down the lightning version. I know it is unlikely but the `^` could give us the situation where we test `8.6.1` in PR, Drupal then release `8.6.2` before the PR gets merged and we actually end up performing a completely untested upgrade of Drupal in production.

LightGuard: Yeah, I can lock down core Drupal and the rest. We should *probably* do the same for the dev ones and lock them at a treeish too.

LightGuard:
I double checked the two errors, they're false positives.

robpblake:
retest this please

LightGuard:
@robpblake good now?

PR #2606, opened 2018-09-13
[DEVELOPER-5364] Determine changed content between runs of the export

robpblake:
This PR adds the ability for the export process to determine which page content has changed between runs. It does this by inspecting the current `sitemap.xml` and comparing the `<lastmod>` entry for each page against that from the previous export.

It locates the previous export by looking in the export archive directory. Each time an export is successfully rsynced, a copy is made into the archive directory to facilitate easy rollback. This copy contains the previous version of the `sitemap.xml` with previous content modification times.

This PR merely adds the ability to detect which pages have changed, it does not add anything to act on that e.g. clearing the Akamai cache. This will be added in another PR.

Additionally this PR removes an unnecessary call to clear the Drupal cache each time an export was run.

### JIRA Issue Link
* [DEVELOPER-5364](https://issues.jboss.org/browse/DEVELOPER-5364)

### Verification Process

* Ensure that the build completes successfully
* Ensure that the tests pass against the exported version of the site

### DIFF
--- a/_docker/lib/export/export.rb
+++ b/_docker/lib/export/export.rb
@@ -8,6 +8,7 @@ require_relative 'export_inspector'
 require_relative 'export_archiver'
 require_relative 'export_archive_pruner'
 require_relative 'export_html_post_processor'
+require_relative 'export_diff'
 
 #
 # Class that acts as a controller of the export process from Drupal
@@ -16,7 +17,7 @@ require_relative 'export_html_post_processor'
 #
 class Export
 
-  def initialize(drupal_host, export_directory, cron_invoker, page_url_list_generator, export_strategy, rsync_handler, rsync_destination)
+  def initialize(drupal_host, export_directory, cron_invoker, page_url_list_generator, export_strategy, rsync_handler, rsync_destination, export_diff)
 
     @log = DefaultLogger.logger
     @drupal_host = drupal_host
@@ -26,7 +27,7 @@ class Export
     @export_strategy = export_strategy
     @rsync_handler = rsync_handler
     @rsync_destination = rsync_destination
-
+    @export_diff = export_diff
   end
 
   #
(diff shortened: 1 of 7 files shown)


robpblake:
@LightGuard Would you mind casting your eye over this one please. I'm merging into `stage` branch at the minute so that I can test this in staging without impacting production.

cheers,

Rob

_docker/lib/export/export_diff.rb
```diff
@@ -0,0 +1,124 @@
...
+            @log.info("URL '#{ url }' did not exist in the previous export and is new content. No cache clear required.")
+            next
+          end
+
+          if Date.parse(current_timestamp) > Date.parse(previous_timestamp)
+            @log.info("Content of URL '#{ url }' has changed. <lastmod> in current sitemap '#{ current_timestamp }', <lastmod> in previous sitemap '#{ previous_timestamp }'.")
+            modified_urls.push(url)
+          end
+
+        end
```
LightGuard: Feels like there should be some sort of faster way of doing this using some set theory, but my brain isn't working enough to figure that out right now.

_docker/lib/export/export.rb
```diff
@@ -42,6 +43,7 @@ def export!
 
     if !@rsync_destination.nil? and !@rsync_destination.empty?
       @rsync_handler.rsync_static_export(export_directory, @rsync_destination, true)
+      @export_diff.modified_content_urls(export_directory)
```
LightGuard: I'm also wondering if we need to do a full export if we already know the diffs. Can we simply feed httrack a list of URLs that are we need to export after we've diffed the sitemap.xml files?

robpblake: @LightGuard The honest answer is "I don't know". 

I'll pull that idea out as another potential improvement to the export process as it feels beyond the scope of this story and it would involve further re-working of the export code.

I've created [DEVELOPER-5381](https://issues.jboss.org/browse/DEVELOPER-5381) to spike that issue.

cheers,

Rob

PR #2609, opened 2018-09-17
DEVELOPER-5373-5328-FontAwesome icon fix

alistairmc:
### JIRA Issue Link
https://issues.jboss.org/browse/DEVELOPER-5328
https://issues.jboss.org/browse/DEVELOPER-5373


KamiQuasi:
@alistairmc we'll need to update the tests to include the icons too, then we can go ahead and merge

alistairmc:
@KamiQuasi  I can't get this to work, its added the test changes but the content has not changed.

alistairmc:
@KamiQuasi  The Problem was I had already added the icons to the test, only seems to pass if I don't have the icons in the test, the test page seems to not have icons for some reason? very odd

LightGuard:
@alistairmc and @KamiQuasi, is this ready to merge?

alistairmc:
Fine from my perspective not sure why the test don't seem to have the right code, but they pass and the rendering is correct in the site. @KamiQuasi  ok from your perspective?

PR #2616, opened 2018-09-18
Moving some images and fixing CSS display

LightGuard:
### JIRA Issue Link
* Not sure if there is one yet

### Verification Process
The background on Microservices should have the design on it instead of a flat color


robpblake:
@LightGuard I'm not sure what I'm looking for here. Do you have a screenshot of what is should look like?

LightGuard:
Grr.... I used an absolute link in the CSS so in the export for the PR it won't show.

robpblake:
@LightGuard I can verify that the file is included at the export where it should be:

https://developers-pr.stage.redhat.com/pr/2616/export/images/Blue_Background@2x.png

PR #2618, opened 2018-09-20
[DEVELOPER-5365] Flush Akamai Cache of pages that have changed content between export runs

robpblake:
### Overview

This PR updates the export process to invalidate entries within the Akamai cache for pages that have changed content between the current and previous run of the export process. By invalidating the entry in the cache, the content is fetched again from the origin server thereby reflecting more-quickly the most up-to-date changes.

To invalidate the Akamai Cache we make calls to the [Fast Purge API](https://developer.akamai.com/api/core_features/fast_purge/v3.html#postinvalidateurl) and in particular make use of the `Invalidate by URL` method. We simply provide a list of the URLs that have changed in this request and Akamai will then make the cache entry as stale, causing the content to be fetched again on the next page visit.

The URLs passed to the `invalidate` method are the full `https` URL of the page that a visitor would type into their address bar. 

I also fix a bug in the export comparison process here whereby the compared `sitemap.xml` files were actually the same because we archived the current export before doing the comparison.

#### Authenticating with the Akamai API

The code makes use of the [Akamai-EdgeGrid Ruby Gem](https://rubygems.org/gems/akamai-edgegrid/versions/1.0.6), which provides a small extension to the typical HTTPClient library within Ruby. This allows us to set API authentication information on the client by reading from a `.edgerc` file for the current user. More information can be found [here](https://developer.akamai.com/legacy/introduction/Conf_Client.html).

As the `.edgerc` file contains sensitive information, it will be managed by Puppet and provisioned to the `stage` and `production` VMs in the same way as we manage other secrets. The `.edgerc` file is then mounted into the export Docker container as a bind mount. You'll notice the updates to the `docker-compose.yml` files for the staging and production to reflect this.

#### Enabling/disabling the cache invalidation

The cache invalidation process is quite defensive and will only proceed if there are URLs to flush and a valid `.edgerc` can be found for the user the process is running as.

In addition the invalidation can be enabled/disabled by setting the `drupal.export.attempt_flush_cache` environment variable to either `true` or `false` respectively. By default this is `false` which means that the cache invalidation does not run unless explicitly enabled. This has been added so:

- Cache invalidation can be disabled without changes to the code for any reason
- In environments where there is no Akamai cache, the export process will not attempt a cache invalidation

The configuration for the production export job in Jenkins will be updated to set the `drupal.export.attempt_flush_cache` value to `true`.

### JIRA Issue Link
* [DEVELOPER-5365](https://issues.jboss.org/browse/DEVELOPER-5352)

### Verification Process

As we have no Akamai cache in PR environment, this will have to be manually tested in `stage` before we go live into `production` with it. 

This PR exists primarily for a code review, to ensure that the export process still works and to allow us to merge this change into `stage` for the manual testing phase.


Gemfile
```diff
@@ -45,7 +45,7 @@ gem "octokit", "~> 4.0"
 gem 'docker-api', :require => 'docker'
 gem 'uuid'
 gem 'listen', '3.0.8'
-
+gem 'akamai-edgegrid', '~> 1.0', '>= 1.0.6'
```
LightGuard: Do you want to lock this down to an actual version?

_docker/Gemfile
```diff
@@ -3,6 +3,7 @@ source "https://rubygems.org"
 gem 'gpgme'
 gem 'docker-api', :require => 'docker'
 gem 'typhoeus'
+gem 'akamai-edgegrid', '~> 1.0', '>= 1.0.6'
```
LightGuard: Same here.

Gemfile
LightGuard: Actually wondering if we can trim a bunch of stuff here since we don't use awestruct anymore.

_docker/lib/export/akamai_cache_flush.rb
```diff
@@ -0,0 +1,135 @@
...
+
+      #
+      # Loads the Akamai client using the ~/.edgerc configuration file. Please see the following for more information
+      # on how this works:
+      #
+      # https://developer.akamai.com/legacy/introduction/Conf_Client.html
+      #
+      def akamai_client
+
+        if(!File.exist?(@edgerc))
```
LightGuard: This could easily be turned into an `unless` condition

Gemfile
robpblake: 👍

_docker/Gemfile
robpblake: 👍

_docker/lib/export/akamai_cache_flush.rb
robpblake: 👍

robpblake:
@LightGuard I've made all the changes that you requested.

cheers,

Rob

PR #2620, opened 2018-09-20
[No Issue] [DO NOT MERGE] Testing to see if we can speed up PR exports.

robpblake:
### DON'T MERGE THIS. IT IS A TEST!


LightGuard:
Nope, didn't make any change. I guess that rules out the network. Slow database? Slow disk?

robpblake:
@LightGuard My gut feeling is that it is networking related. I'll close this seeing as it made no difference.

PR #2624, opened 2018-09-24
[DEVELOPER-5365] Invalidate pages in the Akamai Cache that have changed content between exports

robpblake:
This PR follows up on #2618 and merges the now tested changes into `master` branch as they are ready to be used in production.

One small change as a result of the testing. I have updated the `export_diff` code to use `Time.parse()` rather than `Date.parse()` when comparing the timestamp of sitemap.xml entries. Using `Date.parse()` meant that only the date component of the timestamp was compared, so content changed twice on the same day was deemed to have not been modified.

### JIRA Issue Link
* [DEVELOPER-5365](https://issues.jboss.org/browse/DEVELOPER-5365)

### Verification Process

This has been manually tested in `stage` as part of the work done under #2618.


robpblake:
@LightGuard This is now ready to go into `master` so could you please review and merge.

One small change was made as a result of the testing. I've swapped `Date.parse()` to `Time.parse()` when comparing the `sitemap.xml` entries of two exports. Using `Date.parse()` meant that page modified twice on the same day wasn't deemed to have changed the 2nd time as the date component of the timestamp wasn't different.

cheers,

Rob

PR #2633, opened 2018-09-27
[DEVELOPER 5307] remove time from video publish date

alistairmc:
### JIRA Issue Link
https://issues.jboss.org/browse/DEVELOPER-5307

### Verification Process
<!--
What process should a tester use to verify this pull request closes the JIRA ticket?

/videos/youtube/03LeehvuOJE/ 

should have no time next to the published date


alistairmc:
This looks good to me could someone review and merge if ok

_docker/drupal/drupal-filesystem/web/themes/custom/rhdp/templates/video-resource/node--video-resource.html.twig
```diff
@@ -14,7 +14,7 @@
                 {% endif %}
             </li>
             <li>
-                <small>Published: {{ date }}</small>
+                <small>Published: {{ node.createdtime|date('D, m/d/Y') }}</small>
```
LightGuard: There's also a field_video_published_date (I think that's the correct name) you could use. I think we're going to start using that for other things, probably should use it here.

alistairmc:
@LightGuard have changed to field_video_publish_date

PR #2640, opened 2018-09-28
Export configs from drupal-data-lite container.

rogeralloymagnetic:
It looks like the latest version of Drupal changed how the config is exported which seemed to be blockin `drush cim` from importing correctly. These are the configs exported from the `drupal-data-lite` container.


LightGuard:
This looks fine, let's make sure it doesn't break anything in tests (which I can't see happening, but weirder things have happened)

PR #2656, opened 2018-10-10
DEVELOPER-5255: Image Style URLs

jordanpagewhite:
In this commit, in the 5 changed template files, we are passing just the
plain-text string to the  src or a <href> attribute(s). Previously,
many of these attributes were having entire Node objects passed to them,
which resulted in instances where HTML comments, for example, were being
injected into the src or href attributes. In addition to the 4 entity
type + display modes detailed in DEVELOPER-5255:

* Node, Events, Tile
* Node, Product, Featured Tile
* Node, Video Resource, Teaser
* Node, Video Resource, Tile

I also discovered a similar issue with the Assembly, Static Item entity
type + display mode, so I resolved that in this commit as well.

### JIRA Issue Link
* https://issues.jboss.org/browse/DEVELOPER-5255

##### Assembly, Static Item (broken)

assembly--staticitem--broken

Here is an example of the behavior of this bug prior to my changes. You can see all of the whitespace and the HTML comments within the <a> href attribute. You **should not** see either whitespace or HTML comments, or anything other than the actual URL being referenced, within the href or src attributes.

##### Video Resource, Teaser (fixed)

videoresource--teaser--fixed

##### Video Resource, Tile (fixed)

videoresource--tile--fixed

##### Event, Tile

Unfortunately, I could not find an instance of the Event, Tile content type + display mode on the website. I asked in Slack, and it seemed from Luke's response that this component may not be currently in use on the website. I still resolved the issue to safeguard the potential use of this component in the future, but there isn't a good way to verify this if the component is not being used on the website. I suppose we could create an instance of this component on a QA or Test environment for the purpose of testing this ticket if that is desirable.

##### Assembly, Static Item (fixed)

assembly--staticitem--fixed

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/themes/custom/rhdp/rhdp.theme
+++ b/_docker/drupal/drupal-filesystem/web/themes/custom/rhdp/rhdp.theme
@@ -513,3 +513,13 @@ function rhdp_preprocess_assembly(&$variables) {
       $variables['eloqua_assembly_id'] = $eloqua_assembly_id;
     }
 }
+
+function rhdp_preprocess_views_view_grouping__product_groups(&$variables) {
+  // Replace Twig HTML comments and unwanted whitespace from the Twig title
+  // variable.
+  $variables['title'] = preg_replace(
+    '/(<!--(.)+-->|\s)/',
+    '',
+    $variables['title']
+  );
+}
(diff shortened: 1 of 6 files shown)


jordanpagewhite:
I discovered another component with similarly broken attributes. Here are examples of the broken Product Group on the /products page. You can observe the the markup injected into the name and image attributes on the <dp-category> tags in the second screenshot.

##### Product group

viewsgroupingproductgroup--broken

##### Product group with inspector open

viewsgroupingproductgroup--broken--html

Using `preg_replace`, I deleted the unwanted whitespace and Twig HTML comments.

robpblake:
retest this please

PR #2659, opened 2018-10-10
DEVELOPER-5378: Update Book Node Edit Form

jordanpagewhite:
This change allows a draft to be saved without a value for the PDF Link
field. I also verified that I could submit the form without a value for
the ISBN field.

### JIRA Issue Link

* https://issues.jboss.org/browse/DEVELOPER-5378

### Verification Process

Go here:

Relative URL: `/node/add/books`

Do this:

* Fill out all of the required fields with some data
* Check the RHDP hosted book checkbox field
* Do **not** enter a value for the ISBN or PDF Link fields
* Leave the moderation state "Save as" value as Draft (the default)
* Save and submit the node edit form

#### You should see

* The form should submit successfully without errors and should redirect you to the newly create Book node

Example:

developer-5378--successful

#### You should **not** see

developer-5378--errors


robpblake:
retest this please

PR #2660, opened 2018-10-10
DEVELOPER-4502: Text format a target _blank

jordanpagewhite:
This adds the option of adding the target attribute to <a> tags using
the Red Hat Developers Approved HTML text format.

### JIRA Issue Link

* https://issues.jboss.org/browse/DEVELOPER-4502

##### Go here

Relative URL: `/node/add/article`

##### Do this

* Fill out the required fields with any data
* For the Image caption field, for example, enter an anchor tag into the WYSIWYG like: `<a href="#" target="_blank">Test anchor</a>`
** Ensure you are using the Red Hat Developers Approved HTML text format
* Save the node edit form to create the Article

##### You should see this

* You should be redirected to your newly created Article
* If you click on the <a> you just created in the Image caption field, the link should load in a new tab
* If you inspect the element, you should see that the `target="_blank"` is there


LightGuard:
@adelasofia you're the reporter on this JIRA ticket, please review and merge if you're happy with it.

PR #2679, opened 2018-10-17
[DEVELOPER-5422] Fuse Page Connector Images

jordanpagewhite:
This adds a Media (Image) field to the Connector content type to store a
thumbnail to be displayed on the Connectors page. The new field is added
to the Connectors REST Export View, which will be consumed by the DCP.
In turn, the front-end of our application will make an AJAX call to the
DCP and retrieve the relative URL to the image from the DCP endpoint and
serve that on the Connectors Fuse page. In the event that there is not
an image uploaded to a Connector, a fallback image will display from
static.jboss.org.

### JIRA Issue Link
* Link to the JIRA ticket(s) this pull request solves


jordanpagewhite:
Ryszard is okay with these changes. He said he will handle the DCP part once it has been deployed to production, and that the field name I was expecting to retrieve from DCP to our front-end is fine.

PR #2690, opened 2018-10-23
Correct font stack on <span> tags in code snippet assemblies

jordanpagewhite:
This includes the CSS changes necessary to make <span> tags within Highlight.js assembly code snippets display with the monospace, serif font stack. NOTE: There are CSS changes included in rhd.css that are not mine. These changes must have been committed in src files, but not dist files.

### JIRA Issue Link
* https://issues.jboss.org/browse/DEVELOPER-5432

### Verification Process

* Add a code snippet assembly to an Article
* You should notice that all of the text within the code snippet should use the "monospace, serif" font stack, particularly <span> tags which previously inherited the <body>-level font stack.

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/themes/custom/rhdp/rhd-frontend/src/styles/assembly/_assembly.scss
+++ b/_docker/drupal/drupal-filesystem/web/themes/custom/rhdp/rhd-frontend/src/styles/assembly/_assembly.scss
@@ -21,6 +21,7 @@
 @import "content_pull_quote";
 @import "content_call_to_action";
 @import "content_image";
+@import "code_snippet";
 
 .assembly[data-audience] {
   display: none;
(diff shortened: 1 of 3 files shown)


fearlsgroove:
Looks like this might include some unrelated changes

LightGuard:
@fearlsgroove yeah, I'm thinking the same thing.

LightGuard:
@jordanpagewhite can you verify?

jordanpagewhite:
It does. I believe that there were CSS changes committed in src files but not dist files, so these changes were included when I executed `npm run styles` locally to build my changes.

jordanpagewhite:
Here is an example of one of the commits to just src files but not the dist files (rhd.css): https://github.com/redhat-developer/developers.redhat.com/commit/8eb293d62e9236262f1fd873e99f5694d4c10f6a

@LightGuard Is it preferable to include these hanging changes to the dist file in another way, and then make this PR after those changes are made?

jordanpagewhite:
@fearlsgroove @LightGuard I had a conversation with Luke about this. He was unaware that this FE dist file was still being tracked in git. He suggested I remove it from git tracking as many of the other FE dist files are (you can review in the .gitignore at the root of the repo). I have added this file to the .gitignore alongside the other FE dist files and I ran `git rm --cached` to remove it from tracking.

LightGuard:
That's looking more like what I'd expect to see.

jordanpagewhite:
There is remaining work needed to make the old method of managing code snippets (in the Body field) look like the new method (the code snippet assembly type).

PR #2695, opened 2018-10-25
[DEVELOPER-5454]  composer env dependencies

jordanpagewhite:
### JIRA Issue Link
* https://issues.jboss.org/browse/DEVELOPER-5454

### Verification Process

* Build a local development environment using the docker-compose.yml file, for example the 'drupal-dev' environment. Everything should function as it normally does
* Build the production environment using the docker-compose.yml file. If you create a terminal session within the built image, you should see that the /vendor directory does not contain any dev dependencies.


_docker/drupal/Dockerfile
```diff
@@ -3,6 +3,7 @@ MAINTAINER Jason Porter <jporter@redhat.com>
 
 ARG http_proxy
 ARG https_proxy
+ARG composer_profile
```
robpblake: This could be defaulted to `production` i.e. `ARG composer_profile="production"` and then the only place it needs to be overridden is in the `_docker/drupal-dev/docker-compose.yml`

robpblake:
@jordanpagewhite It looks like the Drupal for this PR is failing to boot. I suspect it is because we have `drush` is dev dependencies, but we actually use that during the boot sequence for Drupal.

_docker/drupal/Dockerfile
LightGuard: It looks like he's set it up in the docker-composer files, but either way works. I guess I'd lean more toward being explicit, but it isn't a big deal either way.

jordanpagewhite: I haven't used this environment, but should the drupal-dev-local-dcp also use the 'developer' profile?

jordanpagewhite:
@robpblake Oh, interesting. Yeah, I will have to pull out drush and set it as a regular/not-dev dependency. Thanks for pointing that out.

LightGuard:
Yeah, haven't been able to get the logs, @jordanpagewhite try putting drush in regular and see what happens.

_docker/drupal/Dockerfile
robpblake: @jordanpagewhite Yes.

_docker/drupal/drupal-filesystem/composer.json
```diff
@@ -64,7 +65,6 @@
         "behat/behat": "3.5.0",
         "se/selenium-server-standalone": "2.53.1",
         "composer/composer": "1.7.2",
-        "drush/drush": "9.4.0",
         "drupal/console": "1.8.0"
```
robpblake: @jordanpagewhite @LightGuard Is it also worth us having the Drupal Console as a non-dev dependency or is it better to use either one or the other?

jordanpagewhite: @robpblake I have exclusively used Drupal Console as a local development tool, and I believe I've only seen it used that way by others as well. They would accomplish the same results (when it comes to putting them on a server) for clearing the caches, updb, etc. etc., but Drush is the 'tried and true' tool.

jordanpagewhite:
@robpblake Are there any further improvements I can make to this PR at this point?

robpblake:
@jordanpagewhite Sorry I completely missed your last comment on this PR. Could you fix the merge conflict and then I think we're probably good to go if @LightGuard agrees.

cheers,

Rob

robpblake:
retest this please

robpblake:
@jordanpagewhite Looks like there is another merge conflict unfortunately.

I'm also going to be a pain and ask if we can make the production profile the default i.e. unless you tell it otherwise, composer will only run in the production configuration. I should have thought this through earlier, so apologies for wasting your time here. It just makes builds of Drupal on Managed Platform that bit simpler if things default to production mode without any extra configuration.

So could we make the following changes:

`ARG composer_profile=production`

and then reverse the logic check so that it defaults to production mode:

```
RUN if [ "$composer_profile" = "development" ]; \ 
    then /usr/local/bin/composer install --no-interaction; \
    else  /usr/local/bin/composer install --no-interaction --no-dev --optimize-autoloader; \
    fi
```

Sorry for requesting changes at this late stage, I should have thought this through more clearly earlier on.

cheers,

Rob

robpblake:
retest this please

robpblake:
@jordanpagewhite This is nice work! It removes 100Meg from the size of our Drupal Docker image, which is a great saving!

Just to confirm that you're happy with the changes to `composer.lock` in this PR? It's a bit over my head as to why there are so many changes to that given you've only moved the `drush` dependency from `require-dev` to `require`. If you're happy I am.

@LightGuard  Anything else from you? If not, I think this can go in.

cheers,

Rob

PR #2696, opened 2018-10-25
[NO-ISSUE]: DO NOT MERGE: Corrected path for unit-test html report

RubyGemTSL:
### JIRA Issue Link
* NO-ISSUE
### Verification Process
- You can view the js-unit-test-report in jenkins from this PR


RubyGemTSL:
retest this please

RubyGemTSL:
retest this please

RubyGemTSL:
@robpblake @KamiQuasi I've been trying to work on this when I can, but I'm on Holiday with the family at the moment. I've tried to move everything related to fe-unit-tests into the new test directory. It works absolutely fine locally (inside docker), however on CI it doesnt seem to install karma. I had to modify the unit-test job on jenkins in order to test this. I've switched it back to it's original config. in the meantime the unit-tests will continue to claim to pass on git. I will try and do bit in the evening this week, however I can't promise i will finish.

If one of you get a chance could you also take a look?

robpblake:
@RubyGemTSL Can I get some context on what isn't working? Not totally sure what I'm looking at here or where I need to start.

cheers,

Rob

KamiQuasi:
@robpblake the issue that we're trying to sort out is that the report HTML that karma spits out isn't getting created inside the docker container, so we're not getting accurate results displayed in Github that necessarily reflect the Jenkins job.

robpblake:
@RubyGemTSL Can you ping me on Monday about this one please. At the minute we're moving even more stuff into the Drupal web-root which I really don't want to do if we can avoid it. I'm working to keep the Drupal image size & complexity down for our MP migration so I'd like to avoid putting anything in there that doesn't need to be.

PR #2709, opened 2018-11-06
[DEVELOPER-5453]  Keycloak script across environments

jordanpagewhite:
This commit implements a solution that will use a config object,
redhat_developer.keycloak.scriptUrl, that will be defined in the
rhd.settings.yml file within each environment. By default, the Keycloak
library in Drupal will use the Keycloak Prod URL. However, if the config
object scriptUrl value is not the Keycloak Prod URL, then we will update
the library entry with the proper URL from config.

NOTE: This will require adding a scriptUrl key/value in the
rhd.settings.yml file in each environment's rhd.settings.yml file.

### JIRA Issue Link
* https://issues.jboss.org/browse/DEVELOPER-5453

### Verification Process

The following is the location where the keycloak.js should be loaded from:

Environment | Keycloak JS Load Location
-- | --
dev | https://developers.redhat.com/auth/js/keycloak.js
pr | https://developers.redhat.com/auth/js/keycloak.js
staging | https://developers.stage.redhat.com/auth/js/keycloak.js
prod | https://developers.redhat.com/auth/js/keycloak.js


_docker/drupal/drupal-filesystem/web/themes/custom/rhdp/rhdp.theme
```diff
@@ -611,3 +611,31 @@ function rhdp_preprocess_views_view_grouping__product_groups(&$variables) {
...
+
+/**
+ * Implements hook_library_info_alter().
+ *
+ * By default, we will load the Keycloak prod script. If we've specified a
+ * different value in this environment's rhd.settings.yml file, we will alter
+ * that library definition here.
+ */
+function rhdp_library_info_alter(array &$libraries, $extension) {
+  $keycloak_url = \Drupal::config('redhat_developers')->get('keycloak')['scriptUrl'];
```
robpblake: @jordanpagewhite I think we should only do further processing here _if_ the user has provided an override through the the `rhd.settings.yml`, otherwise I think we can safely default to prod Keycloak JS URL in all environments. The only environment where we will override this is in `stage`, everything else will use the prod JS location.

robpblake:
@jordanpagewhite Tests are failing because there is no override provided for the `pr` environment, which means that the keycloak.js is not being loaded correctly I believe.

robpblake:
@jordanpagewhite LGTM! Nice work.

@KamiQuasi Can you give this the once over from a front-end perspective. I see the Keycloak JS being loaded from {{https://developers.redhat.com/auth/js/keycloak.js}} on the site export and login/logout functionality seems to be working as expected on the export.

Login/logout doesn't work on the Drupal instance, but I think that is merely because the prod keycloak doesn't like the URL we're asking to re-direct the user too. I don't see any errors in the developer console about failing to load Keycloak JS anymore.

cheers,

Rob

PR #2713, opened 2018-11-08
[DEVELOPER-5385] Add url decode loop and mobile message

alistairmc:
### JIRA Issue Link

https://issues.jboss.org/browse/DEVELOPER-5385

### Verification Process

screen shot 2018-11-09 at 16 17 07
screen shot 2018-11-09 at 16 10 42

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/themes/custom/rhdp/rhd-frontend/src/scripts/@rhd/rhdp-thankyou-page.ts
+++ b/_docker/drupal/drupal-filesystem/web/themes/custom/rhdp/rhd-frontend/src/scripts/@rhd/rhdp-thankyou-page.ts
@@ -38,7 +38,7 @@ export default class RHDPThankyou extends HTMLElement {
         super();
     }
 
-    template = (strings, name, directLink, recentDownload, boldMsg, normMsg) => {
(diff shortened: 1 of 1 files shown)


LightGuard:
Looks like we're waiting for confirmation. @alistairmc I think some screenshots here would be useful as this isn't the easiest thing to setup to test.

alistairmc:
@alistairmc  You are not kidding! will screenshot before I request any merge

sjcox-rh:
@alistairmc This would only be the messaging for mobile, correct?

sjcox-rh:
@alistairmc Since this is the experience for both desktop and mobile, how does the user download on a desktop? I would think that the inline notification for the mobile download would only appear on mobile based screen sizes. 

I feel like there is a CTA missing for desktop downloading and it should be under the download item name.

alistairmc:
@sjcox-rh the second screenshot is desktop, its the existing functionality. The file will already be downloading when that screen is shown, the direct link is just a backup

alistairmc:
@sjcox-rh the mobile message will only show on mobile devices

sjcox-rh:
@alistairmc  Oh got it, sorry I double checked and now I understand the flow. Yes, this all looks good from a UX perspective.

Thanks!

PR #2715, opened 2018-11-12
Fix Keycloak Rewrite SRC Removal

jordanpagewhite:
The rewrite_keycloak_src() function was removed in PR 2709. The Keycloak
Src is modified/broken at some point, and this method fixes it, so I
will add it back.

### JIRA Issue Link
* DEVELOPER-5453

### Verification Process

Keycloak works again on Prod.

### DIFF
--- a/_docker/lib/export/export_html_post_processor.rb
+++ b/_docker/lib/export/export_html_post_processor.rb
@@ -231,6 +231,8 @@ class ExportHtmlPostProcessor
 
   end
 
+
+
   #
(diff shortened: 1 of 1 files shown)


jordanpagewhite:
I tested this locally, and keycloak.js is loaded successfully from `https://developers.redhat.com/auth/js/keycloak.js`. So, it seems that this will allow Keycloak to load properly in Drupal environments using the absolute path, but the relative path will still be used on the export, due to the rewrite_keycloak_src() method.

LightGuard:
retest this please

PR #2719, opened 2018-11-15
Remove Gulpfile.js in root and from package.json

jordanpagewhite:
This removes the Gulpfile.js from the root of the repo. This is a relic
of an old, deprecated front-end build process. The new front-end build
process uses npm within the rhd-frontend directory within the rhdp
theme. This also removes a reference to the Gulpfile in the package.json
file in the root of the repo.

### JIRA Issue Link
* n/a

### Verification Process

* You can still run the front-end build process successfully from the rhd-frontend directory with npm


RubyGemTSL:
I thought about removing this too in my PR yesterday, do we need the `package.json` in the route also @KamiQuasi?  or can we remove that?

KamiQuasi:
@RubyGemTSL if the test scripts don't need it we could remove it. I thought that file was used as part of the build to then fire off the tests down in the theme folder though.

RubyGemTSL:
@KamiQuasi nope I only left it in there so you could run the tests from the root of the project, but I assume you FE guys spent the majority of your time working in the frontend directory so we should just remove package files from the root. WDYT @jordanpagewhite ?

robpblake:
@jordanpagewhite @RubyGemTSL @KamiQuasi What's the outcome of the above discussion? This one good to go in?

cheers,

Rob

jordanpagewhite:
I will pick this back up on Monday or so, once I am past this Product experience work. This is low priority when compared to the Product experience epic.

LightGuard:
I'm assuming we're still waiting for something here?

jordanpagewhite:
### Updates

This removes the package.json and package-lock.json files from the root
of the repo. As discussed in the PR, on Github, the package.json file is
no longer being used. If that is the case, then there is no point in
keeping the package-lock.json file.

This also implies we _should_ be able to remove the node_modules
directory from the root of the repo.

EDIT: The node_modules directory is not tracked. I should have said that individual developers could delete it from their local environments, not that it could be removed from the repo (since it doesn't exist in the repo).

### Next steps

* Verify that all tests pass and all functionality is still in place after the PR environment rebuilds
* Confirm once more with @KamiQuasi that the FE team is not using package.json or package-lock.json.

PR #2727, opened 2018-11-21
[RHDENG-2643] Removed redirects for DCP integration

robpblake:
This PR remove the redirects within the Drupal VirtualHost configuration that were added for DCP integration. These are now implemented directly at the DCP itself.

### JIRA Issue Link
* https://issues.jboss.org/browse/RHDENG-2643

### Verification Process

* The build should be green


robpblake:
@LightGuard I think you're best placed to give this one the once over. AFAIK all of the redirects are no longer required as the DCP calls the correct URL.

cheers,

Rob

PR #2731, opened 2018-11-26
Feature/developer 5491 v2

alistairmc:
### JIRA Issue Link
fix first part of 
https://issues.jboss.org/browse/DEVELOPER-5491
### Verification Process
<!--
update database publish_at field with created if null
update twig to use published_at

second part will remove field published date


alistairmc:
@LightGuard @robpblake this is the first part of the field copy and twig update will do a part two story if this gets merged

PR #2732, opened 2018-11-27
DO NOT MERGE [DEVELOPER-5520] Assembly featured articles

jordanpagewhite:
### JIRA Issue Link
* https://issues.jboss.org/browse/DEVELOPER-5520

### Verification Process

There should be a Featured Articles assembly type available on the Getting Started content field on Product nodes. This assembly type can be used to reference Article nodes, which will then be presented in the Card view mode. Note that the Getting Started content field is only displayed when the boolean field that allows us to use the new Product page designs/assembly-based approach is checked/true, and then it is displayed on the Getting Started page e.g. `products/rhamt/getting-started` where `rhamt` is the Product URL code.


LightGuard:
There are conflicts here, but it looks like we're not delivering this right away anyway?

jordanpagewhite:
### Updates

It was discussed that this component will not be included in the MVP for the Product experience epic. I will pick this ticket back up next week.

### Next Steps

* Pick work back up next week, and get it to a place where UX can review and FE work can begin.

LightGuard:
@jordanpagewhite conflicts

PR #2737, opened 2018-11-29
[DEVELOPER-5550] Consistent DB names in Dev and PR environments

robpblake:
The outage this morning was caused by a DB migration accidentally targeting the wrong DB name. This worked in `dev` and `pr` environments because the database was named correctly for the migration. However once it got to `prod` the migration failed because the DB there has a different name.

This PR ensures that all environments use a consistent DB name so that any migrations not targeting the correct database should fail in PR environment

### JIRA Issue Link
* https://issues.jboss.org/browse/DEVELOPER-5550

### Verification Process

* This build should go completely green
* Access the Drupal interface and see that it's connected to the `rhd_mysql` database


LightGuard:
Good to merge?

PR #2739, opened 2018-11-29
[DEVELOPER-5502] Updates to dynamic content list to support product page use case: New…

fearlsgroove:
Updates to the dynamic content list assembly to support the product resources use case. 

* Reworks the visual styles of the dynamic content feed based on the mocks.
* Enables articles in the dynamic content and dynamic content feed assemblies.
* Enables more vocabularies on the dynamic content feed to support pulling content related to products (product and product line purpose attributes, tags)
* Adds a visual style to support larger heading format ("Big heading")

An important note: this approach enables a few new vocabularies as filters on the dynamic content list assembly, but it is not prescriptive in what taxonomy strategy should actually be used to identify content related to products. My assumption is that the product purchase attribute is the best fit long term, but not content currently uses that term, so content managers will have to backfill. 

In an effort to ease the transition, Drupal tags and topics are also enabled, but these are not likely to be preferred moving forward as a means of classifying which content is relevant to a product.

Images are not assigned on some articles -- when those articles are pulled into a dynamic content list, the space for the image will be empty. Backfill images as needed to avoid weirdness.

<strong>Missing work:</strong> This PR does not solve the "video list" sidebar requirement of the issue.

### JIRA Issue Link
https://issues.jboss.org/browse/DEVELOPER-5502

### Verification Process
* Visit the homepage, kubernetes, service mesh etc to validate the new format for the dynamic content feed looks good.
* Identify wordpress content that has tags appropriate to pull content related to a specific product, and note the tags.
* Identify drupal content that should be included in a sample product resources assembly, and note and/or update the terms involved. Content authors can use the product or product line purpose attributes, tags, or topics.
* On a sample page (a product page with the product content PR available, or any landing page simply as a demo): 
  * create a new dynamic content list (or edit an existing one)
  * Validate the visual style "Big Heading" Indeed makes the headings big
  * Select filters from the drupal term filter and wordpress tags to select which content should appear
  * Complete any other items in the assembly
  * Save all that and validate that the assembly & page look as expected.
* Author bios are inheriting the new article format. View a few author pages to verify those are acceptable.

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/config/sync/assembly.assembly_type.dynamic_content_list.yml
+++ b/_docker/drupal/drupal-filesystem/web/config/sync/assembly.assembly_type.dynamic_content_list.yml
@@ -5,5 +5,5 @@ dependencies: {  }
 id: dynamic_content_list
 label: 'Dynamic Content List'
 description: 'A dynamically generated list of content from Wordpress and Drupal, along with Disqus comments and additional links.'
-visual_styles: "no-padding-top|No padding top|Remove top padding from this assembly\r\nno-padding-bottom|No padding-bottom|Remove bottom padding from this assembly"
+visual_styles: "no-padding-top|No padding top|Remove top padding from this assembly\r\nno-padding-bottom|No padding-bottom|Remove bottom padding from this assembly\r\nbig-heading|Big Heading|Show a large heading for this assembly"
 new_revision: true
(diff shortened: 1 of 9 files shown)


KamiQuasi:
this is okay to test

KamiQuasi:
this is ok to test

sjcox-rh:
@fearlsgroove This is looking really good. I think this is good to merge. There are a couple small changes but nothing that affect the MVP. We can address them on Monday when work is done towards the video list sidebar.

Here is a doc with the feedback. If you have time today and think you can update this to reflect these changes than feel free but like I said, as far as MVP, this can wait until Monday.

https://docs.google.com/document/d/1cqG1u4Tapx2FYjzfWz1iap-zLWppY0KW3Er82PLSxuM/edit?usp=sharing

PR #2749, opened 2018-12-04
Featured Articles Assembly Type

jordanpagewhite:
This creates a Featured Articles assembly type and allows this assembly
type to be referenced from the Content, Downloads Page and Getting
Started Page fields on the Product content type. The Featured Articles
assembly type allows an editor to reference existing Articles to create
a collection of articles that are pertinent to the page/node that the
Featured Articles assembly is placed upon.

I have also created a custom template for node--article--card.html.twig
and field--assembly--field-articles-featured-articles.html.twig to start
the templating/layout for the FE work. The remaining FE work includes
any styling required to meet the InVision mockups (attached to JIRA
ticket).

### JIRA Issue Link
* https://issues.jboss.org/browse/DEVELOPER-5520

### Verification Process

* There is a Featured Articles assembly type that can be referenced from the Content, Downloads and Getting Started fields of the Product content type
* The Featured Articles assembly type allows editors to reference an unlimited number of articles
* The Featured Articles assembly type displays using the Card view mode of the Article content type
* The following field(s) are presented in the templating layer:

1. Article Title (They should be able to overwrite the title if they want to display a different title on the card)
2. Type of article (i.e. how-to, opinion, etc.)
3. Short description of article
4. Author and author's avatar
5. Published date
6. Last updated date

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/config/sync/assembly.assembly_type.featured_articles.yml
+++ b/_docker/drupal/drupal-filesystem/web/config/sync/assembly.assembly_type.featured_articles.yml
@@ -1,0 +1,9 @@
+uuid: 3f21afe9-f690-4208-bd91-1092a40051cc
+langcode: en
+status: true
+dependencies: {  }
+id: featured_articles
+label: 'Featured Articles'
+description: ''
+visual_styles: ''
+new_revision: true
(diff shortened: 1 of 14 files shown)


jordanpagewhite:
h3. Remaining tasks

1. Article Title (They should be able to overwrite the title if they want to display a different title on the card)

h3. Comments

I do not think that there is an ideal way that we can provide this using a simple Text (Plain) field. In this scenario, there is no way for an editor to *just* override say the 2nd or 3rd, etc. title. They must always override every title preceding the final title they want to override. This is clearly not an ideal solution.

I will have to keep considering possible methods to satisfy this request in a way that also accommodates a reasonable editorial UX.

sjcox-rh:
@jordanpagewhite  This is good to go! 

UX will explore options for best handling long article titles. For now that feature for overriding titles will be pushed out until further notice.

adelasofia:
retest this please

PR #2751, opened 2018-12-05
Update Docs for mysql 8 local install

alistairmc:
### JIRA Issue Link
add some extra doc info for mysql v8 error

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,13 @@ If there is anything missing in this guide please submit a PR.
 
 ### Install and Setup
 
-There are some requirements for running Drupal locally: php, and some extensions, composer, and optionally mysql/mariadb.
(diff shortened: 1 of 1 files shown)


robpblake:
@alistairmc All of our other environments use MariaDB 10.3. Is there a particular reason to use MySQL 8 over using the MariaDB Docker Container in your local dev environment?

alistairmc:
@robpblake to run the ./setup_local_drupal.sh you have to have mysql installed locally, it still uses the docker one.

simplest way to install mysql on Mac is 'brew install mysql' this seems to install 8.0.12 by default which has compatibility issues with the docker container.

Please advise if there is a better way.

LightGuard:
You can do either. All you need are the mysql client programs, which you can only get by installing the DB on a mac.

PR #2758, opened 2018-12-07
Fix Search form in Drupal

jordanpagewhite:
This should fix the issue with the search form in Drupal. We should
thoroughly test how this change (albeit 1 character) impacts the search
functionality of the exported site.

### JIRA Issue Link
* https://issues.jboss.org/browse/DEVELOPER-5540

### Verification Process

* Search still works exactly as it currently is on the exported site
* The search now works within Drupal, which it currently does not for paths like `/products/fuse/connectors/` or any path that is multiple 'levels' deep

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/themes/custom/rhdp/templates/menu/menu--rhd-navigation.html.twig
+++ b/_docker/drupal/drupal-filesystem/web/themes/custom/rhdp/templates/menu/menu--rhd-navigation.html.twig
@@ -56,7 +56,7 @@
     {% endfor %}
       <li class="rhd-nav-search">
         <div class="rhd-search-wrapper">
-          <form class="search-bar" accept-charset="utf-8" action="./search/" method="get" role="search" type="submit">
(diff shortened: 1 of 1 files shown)


jordanpagewhite:
I have verified that the search functionality works in the drupal-site by entering a test query into the search block at the following routes and pressing the Enter key to perform the search:

`/`, `/search`, `/products/rhamt/download`

All 3 of those routes worked as expected. I tried those 3 so that I could verify multiple 'levels' given the original issue.

jordanpagewhite:
I verified that the search functionality worked in the exported-site-preview by submitting a test query to the search block and pressing the Enter key to perform the search at the following URLs:

`https://developers-pr.stage.redhat.com/pr/2758/export/`, 
`https://developers-pr.stage.redhat.com/pr/2758/export/search`,
`https://developers-pr.stage.redhat.com/pr/2758/export/topics/linux/`,
`https://developers-pr.stage.redhat.com/pr/2758/export/products/openjdk/overview/`

The search executed properly when performed at all of these routes.

PR #2763, opened 2018-12-13
Reworking SOLPs and Eloqua to handle real world use cases

fearlsgroove:
This PR reworks SOLPs to accommodate real world use cases.

* Adds a hero slot to SOLPs so we can add hero assemblies there (rich text with image, CTA form, image hero, video hero)
* Reworking how styles apply to eloqua embed (use built-in drupal routing calls instead of hand crafting urls)
* Allow hiding the native H1/title on SOLPs, in case it's handled by an assembly or other manual content
* Makes visual styles more declarative and less magical for embedded CTA forms (inline, sidebar, etc)
* Hides the RHD Footer on SOLPs

## Verification:

* Make a SOLP, try out heros
* Add a content call to action to articles and verify the inline and sidebar versions work


jordanpagewhite:
✅ Adds a hero slot to SOLPs so we can add hero assemblies there (rich text with image, CTA form, image hero, video hero)
✅ Reworking how styles apply to eloqua embed (use built-in drupal routing calls instead of hand crafting urls)
✅ Allow hiding the native H1/title on SOLPs, in case it's handled by an assembly or other manual content
✅ Makes visual styles more declarative and less magical for embedded CTA forms (inline, sidebar, etc)
✅ Hides the RHD Footer on SOLPs

I verified all of the criteria above passed in the Drupal environment. I will approve this pull request now.

PR #2775, opened 2018-12-21
[RHDENG-2486] Install "Read only mode" module for Drupal

robpblake:
This PR installs the [Read only mode](https://www.drupal.org/project/readonlymode) module for Drupal.

As part of deployments onto Managed Platform, we will be putting Drupal into read-only mode so that content cannot be created whilst the new version of Drupal is deployed. This module prevents new content from being created in Drupal, but still allows the site to function as normal for anonymous visitors so that the site can continue to function whilst a deployment is on-going.

### JIRA Issue Link
* [RHDENG-2673](https://issues.jboss.org/browse/RHDENG-2673)

### Verification Process

* The entire build should go green
* Visit `admin/config/development/maintenance` in the Drupal admin interface
* Ensure that there is now an option titled `Read Only Mode`
* Ensure that the Read Only Mode is disabled by default
* Visit `admin/config/development/configuration` in the Drupal admin interface and verify that there are no config changes left to import.


jordanpagewhite:
h3. Notes

With read-only mode enabled, I was able to:

* Authenticate to a user account
  * A record of this authentication is persisted to the database. I am not sure if this is something we should care about. It is worth discussing. If necessary, we could determine a suitable way to disable authentication while read-only mode is enabled.
* Create an Article node as an Admin
  * This seems like a 'default' override provided by the module. I do not necessarily see this as something we need to change, although we should make all admins aware of this.

h3. Maintenance message presented to non-admin users

screen shot 2018-12-27 at 10 37 32 am

jordanpagewhite:
@robpblake This looks great. Lets discuss my note on user authentication (above) when we both return just to be sure that we are alright with the implications of that. Aside from that, I approve of this PR. Thanks!

robpblake:
@jordanpagewhite I don't think it's a problem allowing content editors to authenticate whilst the site is in read-only mode as long as content creation is blocked.

In terms of the admin being able to create content, I'm also OK with that. The message for the site being in maintenance mode is the just the default. Not sure there's any value in changing it at the minute unless someone complains about it.

@LightGuard Any comments from you?

robpblake:
@jordanpagewhite / @LightGuard Would be useful to get this one merged today if at all possible.

cheers,

Rob

robpblake:
@LightGuard Could you merge this one today if you have time and approve?

cheers,

Rob

PR #2776, opened 2018-12-28
[ DEVELOPER 5551 ] add onebox entry

alistairmc:
### JIRA Issue Link
https://issues.jboss.org/browse/DEVELOPER-5551

### Verification Process
check onebox works for node.js as per story


robpblake:
LGTM

PR #2781, opened 2019-01-15
Update Core 8.6.5 and Lightning 3.2.3

jordanpagewhite:
This updates Drupal core to version 8.6.5 and the Acquia Lightning
profile to 3.2.3, including many updates to dependencies of these two
major packages. This should be updated anyway, but I am specifically
working on this to address a 'failure to apply patch(es)' error that Rob
discovered and is blocking his MP work.

### JIRA Issue Link
* n/a

### Verification Process

* Drupal should be on version 8.6.5
* Acquia/lightning should be version 3.2.3
* Executing `composer install` should complete without any error/warning about a patch that could not apply
* Tests should pass


robpblake:
retest this please

PR #2796, opened 2019-01-24
[RHDENG-2815] Only cache GET requests to WordPress blog content

robpblake:
This PR updates the `DrupalLaravelAdapter.php` implementation to only cache URLs that are related to content on the WordPress blog. This is so that all other HTTP requests handled by Guzzle are not cached (particularly those related to SSO).

### JIRA Issue Link
* https://issues.jboss.org/browse/RHDENG-2815

### Verification Process

* Not sure

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/modules/custom/rhd_guzzle_throttle/src/DrupalLaravelAdapter.php
+++ b/_docker/drupal/drupal-filesystem/web/modules/custom/rhd_guzzle_throttle/src/DrupalLaravelAdapter.php
@@ -9,6 +9,7 @@
 namespace Drupal\rhd_guzzle_throttle;
 
 use DateTime;
+use Drupal\Core\Site\Settings;
 use function GuzzleHttp\Psr7\copy_to_string;
 use function GuzzleHttp\Psr7\stream_for;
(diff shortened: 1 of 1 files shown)


robpblake:
retest this please

robpblake:
@LightGuard / @jordanpagewhite This is the PR to only cache HTTP requests for the blog.

@jordanpagewhite I've extended the work you did to place a guard around every method to check if the URL being requested is from the blog. Would appreciate your feedback as always.

I've placed some logging statements that _should_ only appear in `dev` and `pr` environments. If you both think we need to be more explicit in terms of controlling the logging e.g have a property in `rhd.settings.yml` to turn it on/off then I'm happy to do that.

@LightGuard Can you take a look at this to ensure it makes sense. In particular I notice that this will stop caching of requests to the Download Manager (I don't know if that's bad or not).

LightGuard:
Looks like it's working to me.

LightGuard:
retest this please

robpblake:
Please don't merge this. I need to investigate potential that we're DDOS'ing the blog with image pre-fetching

robpblake:
This requires #2806 before it can be merged.

robpblake:
retest this please

robpblake:
retest this please

robpblake:
retest this please

PR #2810, opened 2019-02-01
DEVELOPER-5633: Sitemap broken for product pages

jordanpagewhite:
This resolves an issue where the 'old', paragraph-based subpages were
being indexed for Product pages using the 'new, assembly-based Product
page design/approach. This commit provides a conditional to the
simple_sitemap hook that creates these entries for custom
subpages/routing for the product pages that will distinguish
appropriately between the product nodes using the paragraphs-based
approach and those using the new, assembly-based approach.

### JIRA Issue Link
* https://issues.jboss.org/browse/DEVELOPER-5633

### Verification Process

* The site export can execute successfully
* All product page scenarios are properly handled by the simple_sitemap

**Case 1: Product node using old, paragraphs-based approach**

/sitemap.xml should display a route for each of the paragraph subpages/subroutes defined on the Node edit form. For example, currently for Openshiftio, we have:

```
<url>
<loc>docker/products/openshiftio/overview</loc>
<lastmod>2018-08-17T23:07:03+00:00</lastmod>
<priority>0.5</priority>
</url>
<url>
<loc>docker/products/openshiftio/hello-world</loc>
<lastmod>2018-08-17T23:07:03+00:00</lastmod>
<priority>0.5</priority>
</url>
<url>
<loc>docker/products/openshiftio/docs-and-apis</loc>
<lastmod>2018-08-17T23:07:03+00:00</lastmod>
<priority>0.5</priority>
</url>
<url>
<loc>docker/products/openshiftio/community</loc>
<lastmod>2018-08-17T23:07:03+00:00</lastmod>
<priority>0.5</priority>
</url>
```

_The root is listed as docker here (above) because this was generated from my localhost_

**Case 2: Product node using new, assembly-based approach with referenced assembly for the Content, Download page and Getting Started page assembly fields**

/sitemap.xml should display 3 entries for this product with URLs like:

```
docker/products/openshiftio/overview
docker/products/openshiftio/getting-started
docker/products/openshiftio/download
```

**Case 3: Product node using the new, assembly-based approach with at least one of the following assembly fields being empty: Content, Download page and Getting Started page fields**

/sitemap.xml should contain entries for whichever assembly fields have value(s) and should **not** display an entry for an assembly field that does not have a value.

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/modules/custom/rhd_common/rhd_common.sitemap.inc
+++ b/_docker/drupal/drupal-filesystem/web/modules/custom/rhd_common/rhd_common.sitemap.inc
@@ -39,9 +39,58 @@ function product_node_sitemap_entries(&$links) {
   // this add all the sub pages to the list of links and unsets the one with the node entry.
   $callback = function ($product_node_link, $product_node_link_index) use (&$links) {
     $product = Node::load($links[$product_node_link_index]['meta']['entity_info']['id']);
-    $sub_pages = $product->field_product_pages->referencedEntities();
-    $entries = array_map('build_subpage_link_entry', $sub_pages,
-      array_fill(0, count($sub_pages), $links[$product_node_link_index]));
+
+    // The product page is using the new, assembly-based approach.
+    if ($product->hasField('field_use_new_product_page')
+      && !empty($product->field_use_new_product_page->value)
+      && $product->field_use_new_product_page->value == '1') {
+      /** @var \Drupal\Core\Config\ImmutableConfig $rhd_config */
+      $rhd_drupal_config = \Drupal::config('redhat_developers')->get('drupal');
+      $drupal_env = \Drupal::config('redhat_developers')->get('environment');
+      $url_product_name = $product->field_url_product_name->value;
+      $entries = [];
+
+      $base_url = ($drupal_env === 'pr')
+        ? "http://{$rhd_drupal_config['host']}:{$rhd_drupal_config['port']}"
+        : $rhd_drupal_config['host'];
+
+      // If there is content for the Overview subpage.
+      if (!empty($product->field_content->referencedEntities())) {
+        // Add a sitemap entry.
+        $overview = $links[$product_node_link_index];
+        $overview['path'] = "/products/$url_product_name/overview";
+        $overview['url'] = "$base_url/products/$url_product_name/overview";
+        $overview['alternate_urls']['en'] = "$base_url/products/$url_product_name/overview";
+        $entries[] = $overview;
+      }
+
+      // If there is content for the Getting Started subpage.
+      if (!empty($product->field_getting_started_content->referencedEntities())) {
+        // Add a sitemap entry.
+        $getting_started = $links[$product_node_link_index];
+        $getting_started['path'] = "/products/$url_product_name/getting-started";
+        $getting_started['url'] = "$base_url/products/$url_product_name/getting-started";
+        $getting_started['alternate_urls']['en'] = "$base_url/products/$url_product_name/getting-started";
+        $entries[] = $getting_started;
+      }
+
+      // If there is content for the Download subpage.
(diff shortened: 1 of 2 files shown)


robpblake:
retest this please

_docker/drupal/drupal-filesystem/web/modules/custom/rhd_common/rhd_common.sitemap.inc
```diff
@@ -39,9 +39,58 @@ function product_node_sitemap_entries(&$links) {
...
-    $sub_pages = $product->field_product_pages->referencedEntities();
-    $entries = array_map('build_subpage_link_entry', $sub_pages,
-      array_fill(0, count($sub_pages), $links[$product_node_link_index]));
+
+    // The product page is using the new, assembly-based approach.
+    if ($product->hasField('field_use_new_product_page')
+      && !empty($product->field_use_new_product_page->value)
+      && $product->field_use_new_product_page->value == '1') {
+      /** @var \Drupal\Core\Config\ImmutableConfig $rhd_config */
+      $rhd_drupal_config = \Drupal::config('redhat_developers')->get('drupal');
```
robpblake: @jordanpagewhite I'm not totally up on variable scoping in PHP lambdas, but it seems to me we only need to generate the base URL once per cron run?

_docker/drupal/drupal-filesystem/web/modules/custom/rhd_common/rhd_common.sitemap.inc
```diff
@@ -39,9 +39,58 @@ function product_node_sitemap_entries(&$links) {
   // this add all the sub pages to the list of links and unsets the one with the node entry.
   $callback = function ($product_node_link, $product_node_link_index) use (&$links) {
     $product = Node::load($links[$product_node_link_index]['meta']['entity_info']['id']);
```
robpblake: @jordanpagewhite This is going to introduce a `N+1` selects problem e.g:

1. select the list of products
2. for each product, do at least another one select

This is a data access anti-pattern and seems to be impacting the performance of Cron pretty badly on your PR. More info on the n+1 select [here](https://bojanv91.github.io/posts/2018/06/select-n-1-problem).

I can see what you're trying to do from the code, so I'm wondering if there is a way we can query for each of the different product types e.g.:

1. select all product pages that are of the new type and details of their sub-pages in one query and add to sitemap.xml
   1. e.g "select from p from Product p where p.use_new_product_page=true`
   2. Is there a way we can include the logic to work out if it has the required subpages as well? 
2. select all product pages that are of the old type in one query and then add them to the `sitemap.xml`

robpblake: @jordanpagewhite Alternatively is is possible to a "pre-save" hook on the new product nodes that adds additional fields about the sub-pages of the node e.g. `hasDownloads`? This is a `boolean` field that is set to true when the product page is saved and the `field_downloads_page_content` collection is not empty. That means your query simply then has to load the product list and then look at the boolean fields for each entry in the list as to whether or not it should add additional entries to the `sitemap.xml`.

robpblake:
> The Ruby script should be doing this already

@LightGuard We seem to have a problem where cron is taking a long time to run in PR envs for the `sitemap.xml` generation. The ruby call to generate the `sitemap.xml` has a read timeout of 3 mins and we're regularly exceeding that on PR builds. The problem this is causing is that the `sitemap.xml` generation is not complete by the time the export process starts and as such we're getting missing pages.

Adding this cron run here ensures that the `sitemap.xml` is up-to-date before the export runs and only impacts PR environments (this doesn't seem to be a problem elsewhere).

jordanpagewhite:
This is passing now without the export error

jordanpagewhite:
I verified that the codeready sitemap.xml entries match and are accurate on both the Drupal PR environment and the exported site preview

PR #2811, opened 2019-02-04
[DEVELOPER-5622] Product Category View Undefined Index Base

jordanpagewhite:
This error was a result of a missing handler (relationship) on the
product_groups View, which seemed to be some past image field. I removed
the relationship from the View and verified that the display of the View
remained the same. It did. I also verified that the error no longer was
logged. It was not.

### JIRA Issue Link
* https://issues.jboss.org/browse/DEVELOPER-5622

### Verification Process

Go here:

`/products`

and here:

`/product-testing`

Do this:

* Log in to your Drupal account and go to the logs `/admin/reports/dblog` and ensure there are no errors like `Notice: Undefined index: base in Drupal\views\Plugin\…` logged within the time you visited the two routes above.

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/config/sync/views.view.product_groups.yml
+++ b/_docker/drupal/drupal-filesystem/web/config/sync/views.view.product_groups.yml
@@ -511,15 +511,6 @@ display:
           required: true
           entity_type: taxonomy_term
           plugin_id: entity_reverse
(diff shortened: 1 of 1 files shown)


robpblake:
Following @jordanpagewhite instructions above and confirmed that there are no errors in the logs.

PR #2815, opened 2019-02-06
OpenID Path Logger Variables

jordanpagewhite:
This updates the way that variables are injected into the Logger
messages in a patch for the OpenID Connect module to the standard Drupal
way. This will ensure the output is HTML-safe.

### JIRA Issue Link
* https://issues.jboss.org/browse/DEVELOPER-5648

### Verification Process

* Patch applies cleanly / all build steps are green
* Messages are still logged properly in Drupal when a user logs in via OpenID Connect


robpblake:
retest this please

PR #2816, opened 2019-02-06
[DEVELOPER-4469] add new view to admin

alistairmc:
### JIRA Issue Link
https://issues.jboss.org/browse/DEVELOPER-4469

### Verification Process
Login to drupal
go to admin views and see if there is a new view 'out of date content'

/admin/content/views/out-of-date-content
/admin/content/views/users-with-out-of-date-content


jordanpagewhite:
### users-with-out-of-date-content
screen shot 2019-02-09 at 8 06 08 am

You can see in the screenshot above that users are repeated in the table with different amounts of outdated content.

### out-of-date-content
screen shot 2019-02-09 at 8 12 48 am

I am not sure that the users of this view will know what we mean by "Revision user ID" or, if they do, how to easily find that. Is there a way we could make this filter more intuitive to editors?

rhdp-jenkins-slave lab4 eng bos redhat com_37816_admin_structure_views_view_out_of_date_content_destination _admin_structure_views

Here, for example, is something we could do if we include a User relationship on the view and add a (User) User: Name (autocomplete) (exposed) exposed filter to the view. We could change the filter label to something better too like "Author of last revision (username)" or anything that makes it clear to the user what the purpose of the field is.

Does this seem like a better UX to you overall?

alistairmc:
@jordanpagewhite  Make sense to me if you know how to do what you are suggesting feel free to implement on this branch, I have already spent way more time on this than it should have taken.

Can pick this up on Monday in you can explain exactly what I need to do.

alistairmc:
@jordanpagewhite I have updated the views, let me know if this is what you were suggesting.

jordanpagewhite:
@alistairmc I will take a look at this at some point today. Thank you for taking a look

PR #2824, opened 2019-02-08
[MWES-2461] Activate seckit, login_security and REST UI modules for Drupal

robpblake:
This PR installs and enables the `seckit`, `login_security` and `rest ui` modules for Drupal. Additionally it provides some default configuration values.

The purpose of enabling this modules is primarily to increase the security of our Drupal instance once it is live on the Internet directly. In particular:

* `login_security` - provides us with some additional security around brute force login attempts and also sends out notifications to admins when suspicious login attempts occur
* `seckit` - provides us with better options for controlling security of things like XSS scripting attacks
* `restui` - gives us a greater oversight of what REST endpoints we're exposing and the authentication for those endpoints. Also allows us to prevent specific HTTP methods against endpoints

### JIRA Issue Link
* https://issues.jboss.org/browse/MWES-2461

### Verification Process

* Ensure that the build goes green
* Ensure that there is no dangling configuration to import on Drupal
* Verify that the DELETE HTTP method is not allowed against Node REST API by visiting `/admin/config/services/rest`
* Under `/admin/config/people/login_security` verify:
  * Tracking time period is `2` hours
  * A User is allowed 10 failed login attempts during that time
  * Attack detection begins after 10 failed login attempts
  * Emails are configured to be sent to certain Drupal admins

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/config/sync/core.extension.yml
+++ b/_docker/drupal/drupal-filesystem/web/config/sync/core.extension.yml
@@ -7,6 +7,7 @@ module:
   assembly: 0
   author: 0
   automated_cron: 0
+  ban: 0
   basewidget: 0
   basic_auth: 0
   blazy: 0
@@ -79,6 +80,7 @@ module:
   lightning_workflow: 0
   link: 0
   linkit: 0
+  login_security: 0
   markdown: 0
   media: 0
   media_entity_instagram: 0
@@ -121,6 +123,7 @@ module:
   schemata: 0
   schemata_json_schema: 0
   search: 0
+  seckit: 0
   serialization: 0
   session_limit: 0
   shortcut: 0
--- a/_docker/drupal/drupal-filesystem/web/config/sync/core.extension.yml
+++ b/_docker/drupal/drupal-filesystem/web/config/sync/core.extension.yml
@@ -113,6 +113,7 @@ module:
   redhat_developers: 0
   redirect: 0
   rest: 0
+  restui: 0
   rhd_assemblies: 0
   rhd_common: 0
   rhd_disqus: 0
--- a/_docker/drupal/drupal-filesystem/composer.json
+++ b/_docker/drupal/drupal-filesystem/composer.json
@@ -60,7 +60,8 @@
         "drupal/readonlymode": "^1.0",
         "drupal/session_limit": "^1.0@beta",
         "drupal/seckit": "1.1",
-        "drupal/login_security": "1.5"
+        "drupal/login_security": "1.5",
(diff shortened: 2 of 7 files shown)


_docker/drupal/drupal-filesystem/web/config/sync/core.extension.yml
```diff
@@ -7,6 +7,7 @@ module:
   assembly: 0
   author: 0
   automated_cron: 0
+  ban: 0
```
jordanpagewhite: @robpblake It looks like these changes will also enable the ban module. Is that intentional?

jordanpagewhite:
:white_check_mark: There is no config to import on the Drupal PR environment

screen shot 2019-02-09 at 7 52 08 am

:white_check_mark: The modules are enabled

screen shot 2019-02-09 at 7 55 50 am

I also navigated to each of the module's configuration forms and then check the logs, and they did not log any errors.

jordanpagewhite:
I approve of this PR pending Rob's response to including the 'ban' module in this PR, as I pointed out in an earlier comment. I don't see any issue with that in particular. I just wanted to ensure that Rob was aware of it.

_docker/drupal/drupal-filesystem/web/config/sync/core.extension.yml
robpblake: @jordanpagewhite It is installed as a dependency of the `login_security` module. Just to confirm that I don't need to explicitly list it in `composer.json` as it should be installed by the `drupal/core` dependency?

cheers,

Rob

robpblake:
@jordanpagewhite I've updated this PR to also include the `rest ui` module. Can you please verify again that you're happy with the changes here.

cheers,

Rob

robpblake:
@LightGuard I've disabled the ability to make `DELETE` HTTP requests against the `Node` REST API. I don't see this being a problem for your Maven plugin as that is only using `POST` and `PATCH` as far as I can see. Can you please confirm you're  OK with this?

cheers,

Rob

_docker/drupal/drupal-filesystem/web/config/sync/core.extension.yml
jordanpagewhite: Ah, okay, that makes sense. And, yes, that is correct that it will be downloaded as a dependency of Drupal core.

jordanpagewhite:
@robpblake These changes look good to me. I just took a look around the Drupal PR environment. I think that disabling the DELETE requests is wise if we are not using it. I'll approve this PR

robpblake:
@LightGuard Any comments from you on this one?

cheers,

Rob

PR #2830, opened 2019-02-12
[DEVELOPER-5411] Update iFramely API key to RHD account key.

rogeralloymagnetic:
### JIRA Issue Link
https://issues.jboss.org/browse/DEVELOPER-5411

### Verification Process
- Add Embedded Content
- View that it embeds correctly
- View source of <iframe> and confirm "key" url parameter is set to "c3cf64c36c378b376d8984711011e196"


robpblake:
@rogeralloymagnetic Do you have an example of a page where this is already being used? Just want to check that it's not trying to load the script over HTTP. Unfortunately the magic of our export process has a nasty habit of re-writing protocol relative (i.e `//`) URLS to their HTTP version because we currently host Drupal on a non-HTTPS endpoint internally.

cheers,

Rob

LightGuard:
retest this please

LightGuard:
@RubyGemTSL With the change for SSO, the current failing test needs to be updated. Is this an instance of use already having the updated test, we just need to merge it into this PR?

_docker/drupal/drupal-filesystem/web/themes/custom/rhdp/templates/assembly/assembly--embedded-content.html.twig
```diff
@@ -2,5 +2,5 @@
   <section{{ attributes.addClass('assembly') }}{{ audience_selection }}>
     <a href="{{ embed_url }}" data-iframely-url></a>
   </section>
-  <script async charset="utf-8" src="//cdn.iframe.ly/embed.js?key=71b6124e790b6d7e8cdcba47b2a69ccb"></script>
```
robpblake: @rogeralloymagnetic Is there any security risk for us having this in the open in a publicly clonable Git repo? Is this key specific to this project and paid for?

cheers,

Rob

rogeralloymagnetic: @robpblake I'm following the iFramely best practices for embed.js as outlined here: 
https://iframely.com/docs/embedjs#add-embed-js-to-your-page

And have setup whitelists for developers.redhat.com & localhost:8888 as outlined here:
https://iframely.com/docs/allow-origins

This is a hashed key specific to this paid account, however it would not be able to run on any other domain. As for the localhost whitelist addition, I could not think of another way to allow us to test it in our local dev environments and assumed that the need to pull the protected docker image would help against this. 

I guess in thinking through it again, it is possible for someone to use the key to make calls from any local environment on port 8888, so maybe it's worth discussing what to whitelist vs what will work in local dev. That being said, we can whitelist a test domain as well but you'd have to push it there to see embeds working properly.

RubyGemTSL:
> @RubyGemTSL With the change for SSO, the current failing test needs to be updated. Is this an instance of use already having the updated test, we just need to merge it into this PR?

@LightGuard i'm not sure how this change could break the drupal admin log in test - will check

RubyGemTSL:
retest thins please

robpblake:
@RubyGemTSL / @rogeralloymagnetic I'm 99% this will need a merge of master to resolve the testing issue. I'm not totally sure why, but I think somewhere down the line a `git push --force` has happened which (not on this PR) that means some peoples repos are missing bits of  the history.

cheers,

Rob

robpblake:
retest this please

rogeralloymagnetic:
@robpblake This passes tests with the latest master merge. Can you go ahead and merge this in so we can close this ticket out?

PR #2838, opened 2019-02-14
[MWES-2463] Integrate Drupal with Akamai Fast Purge API

robpblake:
This PR brings integrates Drupal with the Akamai Fast Purge API allowing content editors to request clears of content in the Akamai Cache directly from Drupal.

Please note this functionality is **disabled** in PR and dev environments.

This PR adds a button to the top of the user navigation window when browsing content in Drupal. The button is titled "Refresh Page in Akamai". Clicking on this button will send a request to Akamai to clear the cache entry for the page being viewed. This mechanism gives content editors a way of immediately requesting changes to published.

**Note:** Until Drupal is running on Managed Platform, content editors are still reliant upon the export process running before their content will appear live on the site. I need to merge this functionality back into our current deployment so that they are in-sync before we go live on Managed Platform. 

By default, purges of content are directed at the `staging` Akamai `domain`. In our production configuration we use Drupal configuration overrides to purges to be sent to the `production` `domain`.

### JIRA Issue Link
* https://issues.jboss.org/browse/MWES-2463

### Verification Process

* Log in to Drupal
  * There should be no dangling configuration changes to import
  * Navigate to `admin/config/akamai/config` in Drupal and ensure:
    * CCU Credentials are stored in `.edgerc` at `/credentials/akamai/edgerc`
    * CCU version is `CCUv2`
    * Clearing action is `Remove`
    * Base Path is `https://developers.redhat.com`
    * Domain is `Staging`
* Navigate to  a content page in Drupal and click on `Refresh Page in Akamai` button
  * Ensure you get an error about not `edgerc` not existing.

### DIFF
--- a/_docker/drupal/drupal-filesystem/composer.json
+++ b/_docker/drupal/drupal-filesystem/composer.json
@@ -61,7 +61,8 @@
         "drupal/session_limit": "^1.0@beta",
         "drupal/seckit": "1.1",
         "drupal/login_security": "1.5",
-        "drupal/restui": "1.16"
+        "drupal/restui": "1.16",
+        "drupal/akamai": "3.0-alpha3"
     },
     "require-dev": {
         "drupal/drupal-extension": "3.4.1",
@@ -139,6 +140,9 @@
           "drupal/openid_connect": {
             "Change Keycloak LoginForm button text": "patches/openid_connect/change-keycloak-button-text.patch",
             "Add logging to openid_connect to support SSO debug": "patches/openid_connect/openid_connect.module-logging.patch"
+          },
+          "drupal/akamai": {
+            "Change text on the button to refresh content in Akamai": "patches/akamai/change-button-text.patch"
           }
         }
     }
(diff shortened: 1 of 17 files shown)


robpblake:
@jordanpagewhite I'd appreciate you giving this the once over seeing as you helped me with the Akamai integration. Can you please verify that all of the changes that you would expect to see are in this PR as well as giving the functionality the once-over.

cheers,

Rob

robpblake:
> Hey Rob. I took a look over this, the code changes and I poked around the PR environment quite a bit, and everything looks good to me. I am 99% sure that this is intended behavior, but I just wanted to run it by another person in case. When you click on the Akamai cache clear button (given the current config of the PR env), I should see this status message, correct? "Path to .edgerc file "/credentials/akamai/edgerc" does not exist!"

@jordanpagewhite Yes that's correct. We can't enable Akamai purging on the legacy Drupal servers because they don't currently support HTTPs and if we do that, Akamai credentials are transmitted in plain text (according to the Akamai plugin documentation)

robpblake:
retest this please

robpblake:
@jordanpagewhite Actually I've mis-understood this. It's to do with setting Akamai credentials in Drupal itself where the warning is if the page is accessed over HTTP. Given that we're only providing the credentials via the `/credentials/akamai/edgerc` file directly on our servers, the I don't think this is a problem. I'll update to ensure that `prod` and `stage` load these files correctly, but I will leave things disabled in the pr environment as we don't want to allow Akamai cache purging from there.

cheers,

Rob

PR #2843, opened 2019-02-19
[NO-ISSUE] Verify download redirect

RubyGemTSL:
Added extra check to verify that we're redirected to the correct environment (hello-world page)


RubyGemTSL:
@alistairmc assigning this one to you as you discovered the issue. This basically checks that on download you are redirected to the correct environment's Hello World page.

PR #2867, opened 2019-03-05
Fix for failing drupal login test 

RubyGemTSL:
Updated the drupal log in spec to use sso log in rather than legacy login.


jordanpagewhite:
retest this please

PR #2877, opened 2019-03-14
[developer 4612] topic links

alistairmc:
### JIRA Issue Link
https://issues.jboss.org/browse/DEVELOPER-4612

### Verification Process
links for topics should be clickable now has the option for an alternative url in the config

in /admin/structure/taxonomy you can supply an alternative topic landing page if used it will override the default page.

once this is live an appropriate content team will need to update the new data


alistairmc:
@jordanpagewhite can you take a look at this, I managed to do it without a theme php override, just using twig.

PR #2879, opened 2019-03-18
Developer 5505 followup

fearlsgroove:
A couple small followup issues from the compose merge:

* Fixes the entity browser widget used for share images on page nodes
* Updates beryllium to fix the width on entity browser modal dialogs


adelasofia:
this is ok to test

PR #2817, opened 2019-02-06
[DEVELOPER 5505] Install Compose Module

rogeralloymagnetic:
### JIRA Issue Link
https://issues.jboss.org/browse/DEVELOPER-5505

PR #2882, opened 2019-03-20
[DEVELOPER-5752] Update Drupal Core to 8.6.13

jordanpagewhite:
This updates Drupal core to version 8.6.13, addressing SA-CORE-2019-004,
and then I ran drush updb locally (no changes required) and exported the
config, which consisted of some boilerplate layout_builder YAML changes.
Note that this commit also includes some files in the components
directory. It seems that some recent commit made changes to the
composer.json dependencies (one of which must generate these files) but
failed to commit the files. If you run `composer install` on
upstream/master, you should also notice that these files will be
generated leaving us in an 'unclean' git state.

### JIRA Issue Link
* https://issues.jboss.org/browse/DEVELOPER-5752

### Verification Process

* Drupal core is updated to version 8.6.13 to address SA-CORE-2019-00


jordanpagewhite:
https://github.com/redhat-developer/developers.redhat.com/pull/2817 brought in Beryllium, which includes select2 as a dependency in its composer.json file which is where these files are coming from in the `components` directory. See: https://github.com/select2/select2

PR #2892, opened 2019-03-28
[DEVELOPER-5764]- [DEVELOPER-5755] format changes to curated links and new bonus content assembly

alistairmc:
### JIRA Issue Link
https://issues.jboss.org/browse/DEVELOPER-5764

https://issues.jboss.org/browse/DEVELOPER-5755

### Verification Process
created a curated link assembly and test view styles

add a new bonus content list to page and test


alistairmc:
@KamiQuasi can you review this from a technical perspective, Gina is looking at it for UX.

I know the test is a fail but that is just the search because of the outage, did not want to rebuild due to their being data that would need re-entering.

alistairmc:
retest this please

PR #2898, opened 2019-04-10
[DEVELOPER-5789] Hide web.config with .htaccess

jordanpagewhite:
Infosec requested, via Lukas, that we remove or hide web.config. I am
adding it to an existing .htaccess rule and will test it on the PR
environment.

### JIRA Issue Link
* https://issues.jboss.org/browse/DEVELOPER-5789

### Verification Process

* You cannot access, view or download web.config by hitting the path/route directly i.e. https://developers.redhat.com/web.config


jordanpagewhite:
This test did not work as I had hoped it would. I will think a bit about this and probably see if I can chat with Rob or Lukas about the ticket tomorrow.

jordanpagewhite:
I think that this is ready to merge pending @robpblake 's approval

PR #2911, opened 2019-04-22
DEVELOPER-5745 - Remove use of old SSO bridge

KamiQuasi:
### JIRA Issue Link
* https://issues.jboss.org/browse/DEVELOPER-5745

### Verification Process
Go to this URL: `/search/?t=openshift&f=type~article` and hover over the links. They should reflect the new SSO bridge using download manager rather than the old one (old one would start with `broker.`)

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/themes/custom/rhdp/rhd-frontend/src/scripts/@rhd/rhdp-search/rhdp-search-result.ts
+++ b/_docker/drupal/drupal-filesystem/web/themes/custom/rhdp/rhd-frontend/src/scripts/@rhd/rhdp-search/rhdp-search-result.ts
@@ -10,7 +10,7 @@ export default class RHDPSearchResult extends HTMLElement {
 
     get url() {
         const stage = window.location.href.indexOf('stage') >= 0 || window.location.href.indexOf('developers') < 0 ? '.stage' : '';
-        return !this.premium ? this._url : `https://broker${stage}.redhat.com/partner/drc/userMapping?redirect=${encodeURIComponent(this._url)}`;
+        return !this.premium ? this._url : `https://developers${stage}.redhat.com/partner/drc/userMapping?redirect=${encodeURIComponent(this._url)}`;
     }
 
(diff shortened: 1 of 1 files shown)


_docker/drupal/drupal-filesystem/web/themes/custom/rhdp/rhd-frontend/src/scripts/@rhd/rhdp-search/rhdp-search-result.ts
```diff
@@ -10,7 +10,7 @@ export default class RHDPSearchResult extends HTMLElement {
 
     get url() {
         const stage = window.location.href.indexOf('stage') >= 0 || window.location.href.indexOf('developers') < 0 ? '.stage' : '';
-        return !this.premium ? this._url : `https://broker${stage}.redhat.com/partner/drc/userMapping?redirect=${encodeURIComponent(this._url)}`;
+        return !this.premium ? this._url : `https://developers${stage}.redhat.com/download-manager/link/3867444?redirect=${encodeURIComponent(this._url)}`;
```
robpblake: @KamiQuasi Is there anyway you can reference the Drupal environment configuration here to determine the URL of the Download Manager that should be used?

Each time we deploy Drupal, the URL of the download manager that it should use is provided in configuration, specifically in `rhd.settings.yml` which I believe is then converted into a Drupal configuration object. I believe there is a way to inject this into JS cc @jordanpagewhite 

cheers,

Rob

KamiQuasi: @robpblake at this point since these web components are hard-coded into the page anyway it probably isn't worth the change in this case, but future state post-Summit when we release the Fusion-based search and wire them into Drupal a bit more I'll throw in a `sso-bridge` attribute for the newer `dp-search-result` component we'll be using that Drupal can just hook right into.

PR #2912, opened 2019-04-22
DEVELOPER-5778 - Updated TOC styles

adelasofia:
### JIRA Issue Link
[DEVELOPER-5778 - Article TOC links need to change color on hover/selection/area](https://issues.jboss.org/browse/DEVELOPER-5778)

- Moved the TOC-specific styles to _table-of-contents.scss
- Updated styles to fit the use cases listed below ⬇️ 

### Verification Process
1. Go to /coderland/serverless/serverless-knative-intro/
2. Check that the following cases are working:
- Use case 1: If the reader hovers over "The Compile Driver" in the TOC, the color should change from #0066cc to #004c98. When the reader selects "The Compile Driver" in the TOC, the scroll behavior will work as it currently does but "The Compile Driver" link on TOC will remain #004c98 upon selection.
- Use case 2: When the reader scrolls to Intro within the article, the "The Compile Driver" link on the TOC color should change from #0066cc to #004c98.
- Use case 3: When the reader scrolls to "Your assignment" within the article, the "The Compile Driver" link on the TOC should return to #0066cc and the "Your assignment" link on the TOC should change to #004c98.


adelasofia:
# UX APPROVED

PR #2939, opened 2019-05-08
[DEVELOPER-5756] Add Topics and Tags to all content types.

rogeralloymagnetic:
### JIRA Issue Link
* https://issues.jboss.org/browse/DEVELOPER-5756

### Verification Process
1. Edit all content types
2. Go to Fields tab and look in "Purpose Attributes" field group
3. See "Topics" is first field
4. See "Control Tags" is last field

### Screenshot
Screen Shot 2019-05-08 at 7 27 27 AM
Screen Shot 2019-05-08 at 7 27 37 AM

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/config/sync/core.entity_form_display.node.article.default.yml
+++ b/_docker/drupal/drupal-filesystem/web/config/sync/core.entity_form_display.node.article.default.yml
@@ -76,6 +76,7 @@ third_party_settings:
         - field_tax_promotion
         - field_tax_region
         - field_tax_stage
+        - field_tags
       parent_name: ''
       weight: 15
       format_type: details
@@ -231,7 +232,7 @@ content:
     region: fields
   field_tags:
     type: entity_reference_autocomplete
-    weight: 12
+    weight: 17
     settings:
       match_operator: CONTAINS
       size: 60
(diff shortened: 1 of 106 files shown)


jordanpagewhite:
The field label of `field_tags` on the `books` content type is `Tags`, as opposed to `Control Tags` which seems to be used by other content types. The field is, otherwise, identical to `field_tags` on other content types, so I do not think that this is particularly an issue. It is just a difference that I wanted to point out.

jordanpagewhite:
@rogeralloymagnetic Should the Sub-Product content type, in this PR, have a Fields Compose region and put fields like the Purpose Attributes in that region? I think that this might be a slightly unique case when compared to other content types on d.r.c because this content type has sparse fields and no assembly reference fields, which would normally populate the default Compose region. Let me know what your thoughts are. I am definitely open to discussing different possibilities. My main concern would be that an editor could be confused if the Purpose Attributes field is located on the Fields tab for every content type except for one.

Yeah, after continuing through the other content types, I actually feel now that the Sub-Product content type should mimic what the Upstream Projects form Compose layout is, where the Purpose Attributes are in the Fields tab. I think that it would be confusing editor UX to make that one exception. The Upstream Projects content type is similar to Sub-Product in that it also does not have an assembly reference field for the default Compose tab. Let me know what you think. I just want to make sure we're providing a consistent experience for the editors. Thanks.

rogeralloymagnetic:
@jordanpagewhite Good catches, thanks! Updates in latest commits.

robpblake:
retest this please

PR #2946, opened 2019-05-15
[DEVELOPER-5763] User Roles Permissions

jordanpagewhite:
This generally revises the permissions for our Drupal user roles. In
many instances, permissions that are not necessary are now revoked from
Content Creator, Content Reviewer and Content Admin role(s). This also
creates new permissions: one for the RHD Disqus custom module and
contributes a patch to create one for the Trailing Slash contrib module.
This allows us to properly set permissions to administer these modules.

### JIRA Issue Link
* https://issues.jboss.org/browse/DEVELOPER-5763

### Verification Process

TBD


robpblake:
retest this please

robpblake:
@jordanpagewhite Excuse my ignorance, but why is the patch needed for the trailing slash?

I was also wondering if this is a good time to introduce a new role with permissions to manage redirects?

cheers,

Rob

jordanpagewhite:
@robpblake I had to add a patch for the Trailing Slash module to create a permission to administer the configuration form in the admin UI. Prior to this patch, there was no permission(s) defined by the trailing slash module. This means that any authenticated user could access the admin config form, which I feel is undesirable. Given the way that we currently utilize user accounts in our Drupal instance, it is less of a 'risk' than it would be (for example) if we let anyone create a user account.

Yeah, that is a good point. It would be good to create a role for redirects with this PR. I will take care of that.

robpblake:
@jordanpagewhite Just checking in on the status of this one:

- What's the best way for us to verify that the new permissions don't break anything for current users?
- Are you still intending to include the role for redirect management in this PR?

cheers,

Rob

jordanpagewhite:
I added in the Redirect Manager in my latest commit. I also created Content Creator and Content Reviewer+Content Admin roles within the PR environment to verify common workflows for these roles. In all of my tests of workflows, they all passed.

PR #2952, opened 2019-05-28
[DEVELOPER-5821] added or / and select on dynamic content list

alistairmc:
### JIRA Issue Link
https://issues.jboss.org/browse/DEVELOPER-5821

### Verification Process
Go to dynamic content list on the homepage in the cms, you should be able to tick an option to enable AND filter logic on the list unselected is an OR filter

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/config/sync/core.entity_form_display.assembly.dynamic_content_list.default.yml
+++ b/_docker/drupal/drupal-filesystem/web/config/sync/core.entity_form_display.assembly.dynamic_content_list.default.yml
@@ -11,6 +11,7 @@ dependencies:
     - field.field.assembly.dynamic_content_list.field_links
     - field.field.assembly.dynamic_content_list.field_navigation_title
     - field.field.assembly.dynamic_content_list.field_title
+    - field.field.assembly.dynamic_content_list.field_wordpress_category_logic
   module:
     - field_layout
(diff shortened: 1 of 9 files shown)


robpblake:
retest this please

alistairmc:
@jordanpagewhite Can you take a look at this, please. Seems to be working, the only problem I can see is when you change to and it works fine however when you change back to or it's not refreshing the query wondering if its a cache issue or something, locally a server restart resolves the issue.

Do you have any ideas?

robpblake:
@alistairmc All queries to Wordpress are cached by a piece of middleware on the Guzzle HTTP client. I'm not 100% on the origins of this, but I believe it was because the Drupal integration ended up getting blocked by Akamai as spam traffic.  Unfortunately the caching mechanism is a bit blunt, so if you need some granular control over it, there might be a bit of work to do here.

See https://issues.jboss.org/browse/DEVELOPER-5615 for a bit more background on this and ping me on Slack if you need to discuss anything.

cheers,

Rob

alistairmc:
@jordanpagewhite Can you review this, please.  @robpblake do you know how long the cache keeps content?

robpblake:
@alistairmc Forever.

alistairmc:
@jordanpagewhite I have updated the rest endpoint it seems to be working properly now.

PR #2953, opened 2019-06-04
[MWES-2920] Run pull request builds on Managed Platform

robpblake:
This PR introduces supporting code for migrating the `d.r.c` pull request pipeline to run on Managed Platform. This PR _**does not**_ change the existing pipeline, so the build for this PR should complete as normal. Any failure on this build should be investigated.

Once on managed platform, PR builds will benefit from:

* No export
* Full HTTPS support 
* Downloads work in PR environments
* Keycloak logins work in PR environments
* Drupal runs on its own domain name at the root of the domain

### JIRA Issue Link
* https://issues.jboss.org/browse/MWES-2920

### Verification Process

* The build should go green

### DIFF
--- a/_tests/e2e/tests/config/wdio.conf.base.js
+++ b/_tests/e2e/tests/config/wdio.conf.base.js
@@ -8,7 +8,7 @@ const testType = typeof commandlineArgs.testType === 'undefined' ? 'export' : co
 const exclude = testType === 'export' ? 'drupal' : 'export';
 
 let host;
-if (commandlineArgs['base-url'].includes('developers-pr') && testType === 'drupal') {
+if (commandlineArgs['base-url'].includes('developers-pr.stage.redhat.com') && testType === 'drupal') {
     const parsedUrl = require('url').parse(commandlineArgs['base-url']);
     const prNumber = parseInt(parsedUrl.pathname.split('/')[2]);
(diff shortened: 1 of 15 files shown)


_tests/e2e/tests/config/wdio.conf.base.js
```diff
@@ -8,7 +8,7 @@ const testType = typeof commandlineArgs.testType === 'undefined' ? 'export' : co
 const exclude = testType === 'export' ? 'drupal' : 'export';
 
 let host;
-if (commandlineArgs['base-url'].includes('developers-pr') && testType === 'drupal') {
+if (commandlineArgs['base-url'].includes('developers-pr.stage.redhat.com') && testType === 'drupal') {
```
LightGuard: I thought we had a different URL

robpblake: @LightGuard Yep from now on URLs for preview environments will be `developer-preview-*.something.paas.redhat.com`. I'm just keeping this here so that should we need to back this out, it should keep working.

robpblake:
@LightGuard This one is good to go in. It won't change the existing pipeline, but I need this stuff in master before I can do the cut-over.

cheers,

Rob

PR #2955, opened 2019-06-06
[MWES-3156] Update Akamai to CCUV3

robpblake:
Akamai are ending support for v2 of their FastPurge API which is what we are currently using. Therefore we need to bump it to use v3. The Akamai Plugin supports v3 out-of-the-box, but I have taken the chance here to update the plugin to the latest version and then ensured that we're using v3 by default.

**_Please do not merge this without co-ordinating with me as I need to update firewall rules and Akamai secrets in our production environment first._**

### JIRA Issue Link
* https://issues.jboss.org/browse/MWES-3156

### Verification Process

* The build should go green
* Check that Akamai is reported at version `8.x-3.0-alpha4` on `admin/reports/updates` in Drupal
* View admin/config/akamai/config in Drupal and ensure that CCU Version is Akamai Client CCUv3

### DIFF
--- a/_docker/drupal/drupal-filesystem/composer.json
+++ b/_docker/drupal/drupal-filesystem/composer.json
@@ -98,7 +98,7 @@
         "drupal/seckit": "1.1",
         "drupal/login_security": "1.5",
         "drupal/restui": "1.16",
-        "drupal/akamai": "3.0-alpha3",
+        "drupal/akamai": "3.0-alpha4",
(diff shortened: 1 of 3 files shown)


robpblake:
@jordanpagewhite Could you please give this the once over. Can you approve if you're happy, but not merge it as I need to make some changes to production environment first to support this.

cheers,

Rob

PR #2980, opened 2019-07-08
[MWES-3338] Update Drupal to use local storage for the Twig Cache

robpblake:
In trying to understand our performance problems on Managed Platform, I was looking at where we are storing our Twig Cache and it turns out we're actually storing it on an NFS mount which is going to be horrendously slow. This update instructs Drupal to store the cache at `/tmp/twig` on the container filesystem. Each environment can then make a decision about how it wants to provide the `/tmp/twig` mount point.

One caveat to this change is that it does make the assumption that Twig templates do not change in your multi-web-head environment. If a Twig template was to change in a multi-head environment, then the changes would only be reflected on one of the nodes in the cluster. We would need to co-ordinate a `drush cr` across all cluster nodes for us to correctly deliver the new Twig template. For us I believe this holds as our deployments are immutable, but it does mean that in the future we have to be aware of anything that would allow editors to make alterations to Twig templates on the fly

There's a whole bunch of discussion around this here: https://www.drupal.org/node/2752961

The thing to be aware of in the above discussions is that they are doing "in-place" updates of their templates and then flushing the Drupal caches. We do not do that. Every single one of our deployments is isolated, so we do not have that problem.

### JIRA Issue Link
* https://issues.jboss.org/browse/MWES-3338

### Verification Process

* Tests should go green
* Log in to the pod backing the Drupal instance for this PR and verify that Twig cache is now at `/tmp/twig`


robpblake:
@jordanpagewhite I'd appreciate your take on this one. 

What I'm doing here is moving the Twig cache onto the local container filesystem rather than running on the NFS mount. So basically I'm configuring the Twig cache to be at `/cache/drupal/twig` on the container filesystem rather than `/var/www/drupal/web/sites/default/files/php/twig`.

The benefit of doing that is that we can mount a Memory-based filesystem at `/cache/drupal/twig` rather than using the NFS-disk for the Twig cache. This should greatly reduce the per-request processing time as we're not doing network hops just to read files.

The only "downside" to this approach is that the Twig cache is no longer shared between all nodes. This will be fine unless we have any workflows that allow editors to edit Twig templates on the fly. I don't think we do this, but just wanted to double check with you.

cheers,

Rob

jordanpagewhite:
That is correct. We do not let editors edit Twig templates. I cannot foresee a scenario in which that would be a desirable workflow for this project. The possible/perceived 'gain' is very low, but the risks would be relatively high.

jordanpagewhite:
@robpblake I visited a number of admin routes and performed some common admin workflows, and I used Chrome debugger to benchmark the load times, and this PR actually seemed consistently slower than for example my Drupal core update PR. Shouldn't this one be considerably faster?

robpblake:
@jordanpagewhite Yep it should absolutely be quicker, so lets chat about what you're seeing.

cheers,

Rob

robpblake:
rebuild this please

robpblake:
retest this please

robpblake:
retest this please

alistairmc:
@robpblake I have reviewed this, feel free to merge if you are happy for Jordan not to have reviewed it.

PR #2996, opened 2019-07-25
[MWES-3393] Clean temporary files from previous deployments as part of boot process

robpblake:
This PR makes a small update to the environment provisioning process to clean-up temporary files from the previous deployments.

### JIRA Issue Link
* https://issues.jboss.org/browse/MWES-3393

### Verification Process

* The build should go green
* [Log into the Drupal Container](https://github.com/redhat-developer/developers.redhat.com/tree/master/previews)
  * Ensure that files within the `/var/www/drupal/web/sites/default/files/js` directory are all showing the same date of creation using `ls`
  * Ensure that files within the `/var/www/drupal/web/sites/default/files/css` directory are all showing the same date of creation using `ls`
  * Ensure that files within the `/var/www/drupal/web/sites/default/files/externals` directory are all showing the same date of creation using `ls`


robpblake:
rebuild this please

robpblake:
rebuild this please

robpblake:
rebuild this please

robpblake:
@jordanpagewhite Could I please get you to take a look at this one?

Essentially all this does is remove the `DRUPAL_ROOT . "/sites/default/files/js"`, `DRUPAL_ROOT . "/sites/default/files/css"` and `DRUPAL_ROOT . "/sites/default/files/externals"` directory as we start-up Drupal so that we're not having to work with all of the temporary files from previous deployments.

I'm 99% Drupal simply regenerates all of those files and will re-create the directories when it starts up, but would be good if you could review and +1.

cheers,

Rob

robpblake:
> Approved as long as Rob reads and agrees with my justification of my approval in the Conversation tab

Agreed. My test steps made the assumption that someone would review this on the same day that the Drupal instance was created. As you correctly point out this is not necessarily going to be the case, so when someone (you) did finally review it, it would make complete sense for us to see new images in the `externals` directory for that day.

Issue #3008, opened 2019-08-08
Reduce/Eliminate calls to JSDelivr via Bootstrap theme


jordanpagewhite:
### Description

There are scenarios now where an absurd number of calls are made to the JSDelivr CDN by the Bootstrap theme. In one stack trace in New Relic, 619 calls are made to JSDelivr to load the /admin/content route.

### Acceptance Criteria

* Ensure Beryllium isn't trying to use a CDN for anything
* Update the Bootstrap theme to the latest version to see if that resolves this issue
* Determine if we can just disable the CDN feature from the Bootstrap theme entirely. Ensure that this doesn't break anything.

jordanpagewhite:
I tried disabling the CDN on the Bootstrap theme's setting page (locally), but the method JsDelivr::processDefinition() is still executed for some reason even though I cannot figure out _why_ you would want that code executed if you aren't pulling from a CDN. I entirely commented out the method, and I noticed an improvement in the load time of a few admin pages I tried.

More testing and investigation should be done here, but this verifies that the easiest possible solution of just setting the Bootstrap theme config to 'no CDN' does not resolve this issue.

jordanpagewhite:
*Commented out body of ProviderManager::processDefinition()*
/admin/content
no cache, clear Drupal cache before each request and hard refresh in Chrome

Load time, TTFB
------------------
5.43, 5.42
4.88, 4.86
5.20, 5.17
5.05, 5.03
5.05, 5.01

*ProviderManager::processDefinition()*
/admin/content
no cache, clear Drupal cache before each request and hard refresh in Chrome

Load time, TTFB
------------------
7.89, 7.85
7.32, 7.29
7.18, 7.14
6.96, 6.92
7.71, 7.69

jordanpagewhite:
*Next steps*

* Determine how we can not execute the code in the Bootstrap theme that made any external calls or unnecessarily caches responses. We do not need to load the actual Bootstrap assets in this theme, and we definitely should not be making any network calls to fetch them nor cache those responses.

robpblake:
This is a priority in my mind. We can throw all the caching in the world at Drupal, but it's the external network calls that are killing performance.

jordanpagewhite:
#3010

robpblake:
#3010 Is merged and will go out tomorrow morning.

PR #3010, opened 2019-08-12
[#3008] Update drupal/bootstrap and set no CDN

jordanpagewhite:
This updates the drupal/bootstrap contrib theme to its latest version,
which is a significant improvement/change. This commit also sets the
config of the theme to declare no CDN, as opposed to the JSDelivr CDN.
Both of these changes are necessary to remove an unpredictably large
number of CDN requests that were being made on every admin request to
the JSDelivr CDN. As noted in the Github issue, this should result in a
significant performance increase in pages that use the admin theme(s).

### Github Issue Link
* https://github.com/redhat-developer/developers.redhat.com/issues/3008

### Verification Process

* No calls are made to the JSDelivr CDN
* The average load time of admin pages improves, as noted in the linked Github issue

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/config/sync/bootstrap.settings.yml
+++ b/_docker/drupal/drupal-filesystem/web/config/sync/bootstrap.settings.yml
@@ -5,7 +5,7 @@ _core:
 favicon:
   mimetype: image/vnd.microsoft.icon
   path: null
-  url: 'https://localhost/themes/contrib/bootstrap/favicon.ico'
+  url: ''
   use_default: true
 features:
   comment_user_picture: true
--- a/_docker/drupal/drupal-filesystem/web/config/sync/bootstrap.settings.yml
+++ b/_docker/drupal/drupal-filesystem/web/config/sync/bootstrap.settings.yml
@@ -5,7 +5,7 @@ _core:
 favicon:
   mimetype: image/vnd.microsoft.icon
   path: null
-  url: ''
+  url: 'https://localhost/themes/contrib/bootstrap/favicon.ico'
   use_default: true
 features:
   comment_user_picture: true
@@ -16,3 +16,6 @@ logo:
   path: null
   url: ''
   use_default: true
+cdn_provider: ''
+table_responsive: '-1'
+table: ''
--- a/_docker/drupal/drupal-filesystem/composer.json
+++ b/_docker/drupal/drupal-filesystem/composer.json
@@ -100,7 +100,7 @@
         "drupal/restui": "1.16",
         "drupal/akamai": "3.0-alpha4",
         "drupal/trailing_slash": "^1.0",
-        "drupal/bootstrap": "3.17",
+        "drupal/bootstrap": "3.20",
         "drupal/compose": "1.x",
         "drupal/beryllium": "1.x",
         "drupal/search_api": "1.11"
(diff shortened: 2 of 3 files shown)


jordanpagewhite:
rebuild this please

robpblake:
rebuild this please

_docker/drupal/drupal-filesystem/web/config/sync/bootstrap.settings.yml
```diff
@@ -5,7 +5,7 @@ _core:
 favicon:
   mimetype: image/vnd.microsoft.icon
   path: null
-  url: ''
+  url: 'https://localhost/themes/contrib/bootstrap/favicon.ico'
```
robpblake: @jordanpagewhite Can we make this relative? `https://localhost` isn't going to work in any of the other environments.

cheers,

Rob

robpblake:
@jordanpagewhite Do you think this is why some of our cache entries were so large. If I'm reading through the bootstrap release notes properly, it looks like they were downloading and caching a whole bunch of stuff?

cheers,

Rob

_docker/drupal/drupal-filesystem/web/config/sync/bootstrap.settings.yml
jordanpagewhite: Thanks Rob for noticing this. This must have been exported as 'default' config in the latest version of Bootstrap and I missed it. The favicon within the admin section (pages that load the admin theme(s)) remains the same. I've removed this config value either way though just because it seems unnecessary.

jordanpagewhite:
@robpblake Yes, this did contribute to the cache entries in some way, although I do not have an actual measurement to demonstrate _how much_. I think that would be relatively difficult to measure because it would depend on the number of requests per deployment I suppose or until the laravel_cache table was flushed, truncated or otherwise the cache was rebuilt/flushed. 

Disclaimers aside though, yes, this was having some impact on our cache implementation, specifically the custom Guzzle Throttle cache though and probably very minimally (if at all) any actual Drupal caching mechanism. It seems that I do not see the entries in the `laravel_cache` table in the latest db image I pulled, but I am almost positive I recall seeing them there in a past image.

jordanpagewhite:
### Comparison of this PR environments performance vs. PR#3002

In both scenarios, I will load the /admin/content route to compare load times. This will only offers a relatively small sample size of course, but if the performance delta is significant enough it should be clear.

#### PR#3010 (this PR)
https://developer-preview-3010.ext.us-west.dc.preprod.paas.redhat.com/admin/content

##### Uncached
I will be executing `drush cr` on the command line of the PR environment before each page load and forcing a hard reload in Chrome before each request.

10.31s
9.63s
10.12s
10.27s
9.95s

Avg ~10.06s

##### Cached

826ms
820ms
805ms
917ms
1.33s

Avg ~940ms

#### PR#3002
https://developer-preview-3002.ext.us-west.dc.preprod.paas.redhat.com/admin/content

##### Uncached
I will be executing `drush cr` on the command line of the PR environment before each page load and forcing a hard reload in Chrome before each request.

23.93s
21.35s
17.90s
19.02s
18.31s

Avg ~20.10s

##### Cached

2.92s
2.31s
2.45s
2.40s
2.54s

Avg ~2.52s

### Results

Uncached improvement: ~2 times faster
Cached improvement: ~2.7 times faster

To reiterate, this performance improvement only applies to pages that load the Bootstrap admin theme (i.e. all admin routes/pages, but not the public-facing, anonymous user pages/routes)

robpblake:
@jordanpagewhite Nice work!

PR #3011, opened 2019-08-14
[MWES-3419] Map uid of current user into Docker containers for local development

robpblake:
This commit updates the local development experience to use the UID of the current user as the UID of the process that runs within the Docker containers within the local development environment. This should hopefully rectify any issues we have with the previous approach that used root to bootstrap the environment. In using root, we typically end up with files that are no longer owned by the current user and therefore are a pain to delete without using privilege escalation such as sudo

### JIRA Issue Link
* https://issues.jboss.org/browse/MWES-3419

### Verification Process

* Start Drupal using the `_docker/drupal/dev/run-drupal.sh` wrapper script. Once the boot process has completed, ensure that all files within the `_docker/drupal/dev/drupal-workspace` directory are owned by your `uid`


robpblake:
rebuild this please

robpblake:
rebuild this please

PR #3017, opened 2019-08-20
[MWES-2871] Make memcached default cache engine in all environments but prod and local dev.

robpblake:
This update makes memcached the default cache engine for all environments other than production and local dev to ensure it works as expected with our deployment workflow.

### JIRA Issue Link
* https://issues.jboss.org/browse/MWES-2871

### Verification Process

* Build should go green
* Visit `/admin/reports/memcache` and ensure that two memache instances are listed


robpblake:
rebuild this please

Issue #3007, opened 2019-08-08
Remove rhd_guzzle_throttle custom module and cache Wordpress API and Download Manager calls with Drupal Cache API


jordanpagewhite:
### Description

The `rhd_guzzle_throttle` custom module caches response data from the Wordpress API into a `laravel_cache` table in the database. We should implement this caching using the tools that Drupal core provides.

### Acceptance Criteria

* Data fetched from the Wordpress API and Download Manager is cached via the [Drupal 8 Cache API](https://api.drupal.org/api/drupal/core!core.api.php/group/cache/8.2.x).
  * It seems that the most clear way to accomplish this is to create cache bins for each of the APIs to provide a separation of concerns.

jordanpagewhite:
This be covered by https://projects.engineering.redhat.com/browse/RHDX-67 now. Closing this issue.

PR #3028, opened 2019-08-28
[DO NOT MERGE][GH-3007] Replace rhd_guzzle_throttle with Drupal caching

jordanpagewhite:
This uninstalls the custom rhd_guzzle_throttle module and replaces the
caching that module provided, for external calls to our Wordpress API
and Download Manager, with a standard Drupal Cache API implementation.
This also includes a database update in rhd_common.update.inc to drop
the laravel_cache table as we no longer will use that and do not
otherwise need it.

### Github Issue Link

* https://github.com/redhat-developer/developers.redhat.com/issues/3007

### Verification Process

* Build is green
* Database update hook, `rhd_common_update_8001()`, executes successfully and `laravel_cache` table is dropped

Upon visiting a freshly built environment, Drupal will make whatever Wordpress API and/or Download Manager call(s) are required to load a given page, and then it will cache the responses using the Drupal Cache API and in the `lightning_cache_default` table. You will notice entries like (cid values):

* download_manager_api:product:openjdk
* wordpress_api:by_category:per_page_3-logic_or-categories_4587
* wordpress_api:categories
* wordpress_api:content_media:574937
* wordpress_api:post:520407

These represent the different types of calls we make to Download Manager and the Wordpress API

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/modules/custom/rhd_assemblies/src/Service/DownloadManagerApi.php
+++ b/_docker/drupal/drupal-filesystem/web/modules/custom/rhd_assemblies/src/Service/DownloadManagerApi.php
@@ -58,22 +58,34 @@ class DownloadManagerApi {
    *   sucessful. Otherwise, returns FALSE.
    */
   public function getContentById($id) {
-    try {
-      $feed_url = $this->apiUrl . $id;
-      $request = $this->client->request('GET', $feed_url);
-      $response = $request->getBody()->getContents();
+    $cid = 'download_manager_api:product:' . $id;
 
-      return json_decode($response);
+    // Serve the cached Download Manager data if available.
+    if ($cache = \Drupal::cache()->get($cid)) {
+      $data = $cache->data;
     }
-    catch (\Exception $e) {
-      \Drupal::logger('rhd_assemblies')->error(
-        "Exception while calling Download Manager API: @message", [
-          '@message' => $e->getMessage(),
-        ]
-      );
+    // If this is uncached or the cache is invalid, fetch fresh data.
+    else {
+      try {
+        $feed_url = $this->apiUrl . $id;
+        $request = $this->client->request('GET', $feed_url);
+        $response = $request->getBody()->getContents();
+        $data = json_decode($response);
+        // Persist this data to the download_manager_api cache bin for an hour.
+        \Drupal::cache()->set($cid, $data, time() + 3600);
+      }
+      catch (\Exception $e) {
+        \Drupal::logger('rhd_assemblies')->error(
+          "Exception while calling Download Manager API: @message", [
+            '@message' => $e->getMessage(),
+          ]
+        );
 
-      return FALSE;
+        return FALSE;
+      }
     }
+
+    return $data;
   }
 
   /**
--- a/_docker/drupal/drupal-filesystem/web/modules/custom/rhd_assemblies/src/Service/WordpressApi.php
+++ b/_docker/drupal/drupal-filesystem/web/modules/custom/rhd_assemblies/src/Service/WordpressApi.php
@@ -48,22 +48,34 @@ class WordpressApi implements RemoteContentApiInterface {
    *   Otherwise, this method return FALSE.
    */
   public function getContentById($id) {
-    try {
-      $feed_url = $this->apiUrl . '/wp-json/wp/v2/posts/' . $id;
-      $request = $this->client->request('GET', $feed_url);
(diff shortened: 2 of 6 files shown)


robpblake:
rebuild this please

_docker/drupal/drupal-filesystem/web/modules/custom/rhd_common/rhd_common.updates.inc
```diff
@@ -0,0 +1,29 @@
+<?php
```
robpblake: @jordanpagewhite Is this the preferred way to do database migrations in Drupal?

If yes then I'm happy to deprecate and ultimately remove the `phinx` DB migration framework that we have (assuming you're on board with that?)

When will this run? Is it applied as part of the `drush updb` command? If yes, I assume the failure of the update will cause the `drush updb` command to fail?

The other question I have is how does Drupal know that this update has been applied? Is there some internal state management somewhere that says `rhd_common_update_8001` has already been done?

_docker/drupal/drupal-filesystem/web/modules/custom/rhd_assemblies/src/Service/DownloadManagerApi.php
```diff
@@ -58,22 +58,34 @@ public function __construct(ClientInterface $client, EntityTypeManager $entity_t
...
-      );
+    // If this is uncached or the cache is invalid, fetch fresh data.
+    else {
+      try {
+        $feed_url = $this->apiUrl . $id;
+        $request = $this->client->request('GET', $feed_url);
+        $response = $request->getBody()->getContents();
+        $data = json_decode($response);
+        // Persist this data to the download_manager_api cache bin for an hour.
+        \Drupal::cache()->set($cid, $data, time() + 3600);
```
robpblake: @jordanpagewhite Could we extract this into a configurable value e.g. `$config['redhat_developers']['cache.ttl']['download_manager'] = <time_in_seconds>` with some sane default if ^ is not set? 

In terms of the actual TTL for this data, can you remind me what this integration is actually used for? Is it simply to get product names? If _yes_ that I'm gonna say we can go for a longer TTL on this one.

robpblake:
rebuild this please

robpblake:
@jordanpagewhite Can we also delete the `_docker/drupal/drupal-filesystem/web/modules/custom/rhd_guzzle_throttle` code from the Git repo as part of this?

_docker/drupal/drupal-filesystem/web/modules/custom/rhd_assemblies/src/Service/WordpressApi.php
```diff
@@ -101,30 +113,42 @@ public function getContentByCategory($categories, $max_results, $select_logic) {
...
-          '@message' => $e->getMessage(),
-        ]
-      );
-
-      return [];
+        // Pass the WP posts returned in $results, and get the processed/formatted
+        // results from getContentCompositeMultiple().
+        $data = $this->getContentCompositeMultiple($results);
+        // Persist this data to the wordpress_api cache bin for an hour.
+        \Drupal::cache()->set($cid, $data, time() + 3600);
```
robpblake: @jordanpagewhite Same question here as for the Download Manager integration. Can we extract the TTL to a configuration value with some sane default applied if it's not set?

robpblake:
@jordanpagewhite Our preview environments have Memcached enabled, I'll have a look to see if Drush lets us inspect individual cache entries because those tables are not going to exist in the DB in the preview environments.

jordanpagewhite:
@robpblake I will submit a follow-up PR to this PR (once it is merged) that just removes the `rhd_guzzle_throttle` directory. I cannot remove it in this PR as Drupal will not succeed in uninstalling the module if the files do not exist.

_docker/drupal/drupal-filesystem/web/modules/custom/rhd_common/rhd_common.updates.inc
robpblake: Thanks for the explanation!

robpblake:
rebuild this please

jordanpagewhite:
rebuild this please

robpblake:
rebuild this please

robpblake:
rebuild this please

robpblake:
@jordanpagewhite Given that we're apparently only 6 weeks away from replacing Wordpress, lets pause this. We can always re-open if needs be.

cheers,

Rob

Issue #3039, opened 2019-09-03
Move Composer patch files to local filesystem


jordanpagewhite:
We need to move our patch files, and the references to the patches in composer.json/composer.lock, to the local filesystem. The network calls fail sometimes in our build process, which causes the entire build to fail.

### Acceptance Criteria

* For any patch that currently depends on a network call to some other site i.e. drupal.org, download the .patch file and place it in the /patches/[module_slug] directory.
  * If a [module_slug] directory does not exist for a given module (or core), create one.
* All patches defined in composer.json should use relative file paths (not network calls to drupal.org or githubusercontent.com etc.)

jordanpagewhite:
How timely! I just ran into this issue as I was creating the issue.

Ex.

```
   Could not apply patch! Skipping. The error was: The "https://www.drupal.org/files/issues/2019-06-21/2174633-309.patch " file could not be downloaded (HTTP/1.1 500 Domain Not Found)
```

jordanpagewhite:
https://github.com/redhat-developer/developers.redhat.com/pull/3040

robpblake:
No longer required as my work to fix the issues with the proxy and also to make use use the Nexus NPM registry seems to have resolved the instability here.

PR #3040, opened 2019-09-03
[DO NOT MERGE][SURGE-3039] Move Composer patches to local filesystem

jordanpagewhite:
This puts all of our patch files into the local filesystem of our repo,
as opposed to some which were hosted on drupal.org and one on github.
This will avoid an issue we were seeing where managed platform could
fail to find some domains.

### Issue Link
* #3039 

### Verification Process

* All Composer patches download from the ./patches/ directory on the filesystem of the repo, as opposed to making network calls to download .patch file(s)


jordanpagewhite:
Okay, that is good to know. Thanks @robpblake . Lets see if #3041 entirely resolves this issue. If so, then we can just close this PR and https://github.com/redhat-developer/developers.redhat.com/issues/3039

robpblake:
I think this is now fixed by the work done in #3041 and also #3043 . I will close this, but if this comes back we may need to re-visit.

Issue #3051, opened 2019-09-05
Code Snippet


jordanpagewhite:
FE Ticket:

https://github.com/redhat-developer/rhd-frontend/issues/143

PR #3071, opened 2019-09-11
[SURGE-3051] Code snippet assembly type

jordanpagewhite:
### Issue Link
* #3051 

### Verification Process

* Drupal component should meet UX requirements in Marvel design: https://marvelapp.com/c6c454h/screen/60311375

Test Drupal page for review: https://developer-preview-3071.ext.us-west.dc.preprod.paas.redhat.com/node/212490/


rogeralloymagnetic:
> These should only expand to however wide the content is within the article (756 px max width) rather than expanding throughout the whole screen

This assembly does expand to the parent container. The demo just doesn't have a wrapper element.

rogeralloymagnetic:
@jordanpagewhite I'm not sure where the conversation ended regarding wrapping this in a `<section>`. Approving for now, but I'll let you merge if you don't want to add it, or update it if you do.

PR #3107, opened 2019-09-25
[MWES-3558] Update local dev to pull data images from PSI Upshift registry

robpblake:
The Docker registry on which we host data images for local development is being decommissioned and therefore we need to migrate to the PSI hosted registry.

### JIRA Issue Link
* https://issues.jboss.org/browse/MWES-2904

### Verification Process

* The build should go green


robpblake:
retest this please

PR #3151, opened 2019-09-30
Add Authors to Ebook and Cheat Sheet Cards

jordanpagewhite:
### JIRA Issue Link
* n/a

### Verification Process

- Go to the homepage
- Scroll down a bit to find Cheat Sheet and Ebook cards and verify that the ones with authors set actually display the authors

Screen Shot 2019-09-30 at 2 27 11 PM

### Known issue(s)

There is an issue with Card spacing that will be resolved in this ticket in the rhd-frontend repo

https://github.com/redhat-developer/rhd-frontend/issues/229

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/themes/custom/rhdp2/components/card.twig
+++ b/_docker/drupal/drupal-filesystem/web/themes/custom/rhdp2/components/card.twig
@@ -18,6 +18,9 @@
  * - content_image: An image rendered within the content of the Card.
  * - comment_count: A string integer of the number of comments made on the
  *   content within the card.
+ * - cta:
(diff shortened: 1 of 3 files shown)


jordanpagewhite:
@sjcox-rh Yeah, of course. No problem. I just added that author to that piece of content on Prod. It will take a day for another database image to be generated and then we would have to rebuild this PR environment. Would you be okay with approving this PR though, seeing that the functionality to display the author(s) is in place and knowing that the author has been added for that piece of content on Prod?

sjcox-rh:
@jordanpagewhite thanks for adding the author. And of course, this is good to merge!

Issue #3120, opened 2019-09-25
node--article--full


jordanpagewhite:
See design in Marvel: https://marvelapp.com/c6c454h/screen/60311375

PR #3154, opened 2019-10-02
[SURGE 3120] Article content type

ghost:
### JIRA Issue Link
* https://github.com/redhat-developer/developers.redhat.com/issues/3120

### Verification Process
1. View article page
2. Match to mockups: https://marvelapp.com/c6c454h/screen/60311375/

##### Article page examples

- https://developer-preview-3154.ext.us-west.dc.preprod.paas.redhat.com/articles/red-hat-developer-program-benefits/
- https://developer-preview-3154.ext.us-west.dc.preprod.paas.redhat.com/articles/renew-your-red-hat-developer-program-subscription/
- https://developer-preview-3154.ext.us-west.dc.preprod.paas.redhat.com/articles/c-containers/
- https://developer-preview-3154.ext.us-west.dc.preprod.paas.redhat.com/rhel8/install-rhel8/


sjcox-rh (changes requested):
@rogeralloymagnetic These are looking good!  I noticed a couple small updates that are needed:

1. In the left hand jump to nav (ToC) there should be 16px padding on all sides and 16px space between each item. I noticed some inconsistencies between articles:
For example, there is too much space between items here
Screen Shot 2019-10-04 at 8 52 23 AM

For this example, there is no padding on right side of table
Screen Shot 2019-10-04 at 8 52 32 AM

2. The social icons needs to be 16px. I noticed there was a 24px bg on each icon. Can that be removed and just use a gray 16px icon? (currently icon is white)

3. Author image is too large. Let's use the same sized circle implementation used across site for authors.
Screen Shot 2019-10-04 at 8 54 52 AM

jordanpagewhite:
@sjcox-rh Re: 1. Here is the same example on the live site:

Screen Shot 2019-10-04 at 10 11 58 AM

The unbroken string 'Configuration/Management' is **very** long relative to its placement. We could just enforce a min-width on _all_ such elements based upon this string, but I don't think that would be desirable. You can see that a padding is actually enforced (or attempting to be enforced) but the long string cannot wrap

Screen Shot 2019-10-04 at 10 13 44 AM

jordanpagewhite:
@sjcox-rh We could do something like this where we wrap in the middle of a word. This is also what it looks like with 16px of padding around the entire ul element and 16px of space between each list item.

Screen Shot 2019-10-04 at 10 19 24 AM

Do you prefer it to wrap in the middle of a word like this?

sjcox-rh:
@jordanpagewhite thanks for explaining what caused this issue. I think we leave it as is for now since this is a very specific use case to handle long strings.  I want to avoid breaking up/hyphenating words. This will just have to be monitored as articles are published and we can update headers so that they fit within the table of contents.

jordanpagewhite:
Okay, cool. Thanks @sjcox-rh. So to summarize remaining feedback for Roger:

1. The social icons needs to be 16px. I noticed there was a 24px bg on each icon. Can that be removed and just use a gray 16px icon? (currently icon is white)

2. Author image is too large. Let's use the same sized circle implementation used across site for authors.

rogeralloymagnetic:
@sjcox-rh I've added back the missing author file which now correctly displays it as an author-tile.

I've updated the social icons as best as I can. It's using the AddToAll widget which only gives us so much flexibility. I can't get rid of the background, but I've made it transparent and 18px in an attempt to get the main icon to be around 16px. Let me know if you want to try a different size on this.
Screen Shot 2019-10-04 at 2 37 36 PM

I've also tightened up the sidebar padding to account for line-height and other elements padding to get the real-world spacing to be 16px everywhere.
Screen Shot 2019-10-04 at 2 40 51 PM

sjcox-rh:
@rogeralloymagnetic great those changes are much better, thank you! 

Also, Im sorry I didn't notice this sooner, but I JUST noticed the behavior or the ToC on mobile and I take back what I said @jordanpagewhite about leaving ToC as is. The current implementation is fine for desktop, but we need to update it to use the same behavior we are currently using on the live site.  At a certain breakpoint, the ToC should stop decreasing it's width and just move above the article content. Since we have more and more users reading articles on mobile, this needs to be updated. 

Current mobile behavior or PR:
Screen Shot 2019-10-06 at 8 46 04 PM

Live site ToC behavior:
Screen Shot 2019-10-06 at 8 48 02 PM

So so sorry I didn't catch this on Friday!!!!

gdoyle1:
@rogeralloymagnetic @sjcox-rh In addition to SJ's comment when we have articles with an image to the right of the title, let's follow the current responsive behavior of it dropping below the title & metadata section as well on breakpoint-md (768px width screen)! 
Screen Shot 2019-10-07 at 7 55 30 AM
Screen Shot 2019-10-07 at 7 55 19 AM

gdoyle1 (changes requested):
See the comment above ^

sjcox-rh:
@rogeralloymagnetic As for when/where regarding the third column (right section with series, etc). That should also break at breakpoint-md (768px width screen) and be placed under the article content.

So order is as follows:

**breakpoint-md (768px width screen)**
- Article title
- Image (if added)
- ToC + article
- Right sidebar (series, etc)

'**breakpoint xsmall: 320px + small: 480px**
- Article Title
- Image (of added)
- ToC
- Article Content
- Right sidebar (series, etc)

Let us know if you would like some mocks created!

sjcox-rh:
@rogeralloymagnetic updated the comment above to clarify large breakpoint as well. @gdoyle1 is going to mockup the various breakpoints and well add the link here for you to reference.

gdoyle1:
Mockups for you @rogeralloymagnetic https://marvelapp.com/c6c454h on screens 6-9

gdoyle1 (changes requested):
@rogeralloymagnetic The only thing I noticed is that the text width doesn't stretch all the way to the gutters on tablet and mobile ... looks like there is double gutters happening

sjcox-rh:
@rogeralloymagnetic YES, this is much better! This is good to merge.

Just FYI, there is a minor change but I'll just open up a follow up ticket to address it. It's to match the padding of the article to the ToC and Image at the med and small/xsmall breakpoints. Right now there seems to be an extra layer of padding applied to the article content so it doesn't align with the ToC.

Screen Shot 2019-10-07 at 11 53 10 AM

rogeralloymagnetic:
@gina & @sjcox-rh I think this is an issue with FE not accounting for an assembly with `component` class nested in a grid. Thanks for making a new ticket for that!

sjcox-rh:
@rogeralloymagnetic okay thanks for letting me know. I moved the issue to FE

Issue #3106, opened 2019-09-24
Collection assembly


jordanpagewhite:
See http://developers.redhat.com/training for example. Card grid with a pager

This assembly type can display the Article, Book and Video Resource node types using the Tile view mode.

rogeralloymagnetic:
PR: https://github.com/redhat-developer/developers.redhat.com/pull/3157

jordanpagewhite:
Closed by #3157

PR #3157, opened 2019-10-03
[SURGE 3106] Collections assembly

rogeralloymagnetic:
### JIRA Issue Link
* https://github.com/redhat-developer/developers.redhat.com/issues/3106

### Verification Process
1. view `/training`
2. Check Collections assembly matches designs


sjcox-rh:
@rogeralloymagnetic @jordanpagewhite not sure if this is more of a FE issue, but when you hover over a title that is only two lines, the card height increases by about 10 px and all content gets shifted down slightly.

Also, can we left align the pagination?

If these are both FE issues, then this is good to merge!

rogeralloymagnetic:
@sjcox-rh The hover issue is FE if you want to make a separate ticket there.

I've updated the pager alignment. 
Screen Shot 2019-10-04 at 3 04 18 PM

Issue #3132, opened 2019-09-25
cta_form assembly


rogeralloymagnetic:
/contact-sales

Estimate: 8

rogeralloymagnetic:
PR: https://github.com/redhat-developer/developers.redhat.com/pull/3186

This technically resolves this ticket as it relates to the sales form, but has work left to do to be production ready for all intended use cases.

PR #3186, opened 2019-10-10
[SURGE 3132] Eloqua cta_form

ghost:
### JIRA Issue Link
* https://github.com/redhat-developer/developers.redhat.com/issues/3132

### Verification Process
1. View https://developer-preview-3186.ext.us-west.dc.preprod.paas.redhat.com/contact-sales/
2. Edit and try all Visual Styles

### Known Issues
1. ~~`.inline` style still needs to be reimplemented for use in Articles~~
2. ~~`content_call_to_action` and `eloqua_form` Assembly Types need to be removed from all Content Types~~
3. ~~`content_call_to_action` should be able to be removed from the system altogether~~


gdoyle1 (approved):
@rogeralloymagnetic I can't get my login working to create an assembly myself so I needed to review based off your example on /contact-sales. Is there a way we can get the interactions in the form fields working as they do in PatternFly (https://www.patternfly.org/v4/documentation/react/components/form/)? We will need to create a follow up issue to address this @jordanpagewhite .

rogeralloymagnetic:
@gdoyle1 We won't be able to easily add in the PatternFly interactions, since this is an Eloqua iFrame. There may be some way to add some of them in an ugly round-about way, but we'd need to set aside time for R&D on it, then decide if it's worth the technical burden that would be required from that assembly.

This is maybe the most complex assembly on the site currently.

CC @jordanpagewhite

gdoyle1:
@rogeralloymagnetic Ah okay - not a blocker so let's just get this one through then! Plus we rarely use this assembly across the site.

jordanpagewhite (approved):
These code changes look good to me. Thanks Roger

PR #3189, opened 2019-10-12
[SURGE-3013] Masthead and navigation

fearlsgroove:
Implements brand, main nav, topic nav, mobile etc.

Keycloak untested but might work. Search is missing from main nav. Needs a few tweaks to rhd-frontend


jordanpagewhite:
the reason search isn't working in the navigation area is because the HTML structure is not the same as what the scripts are expecting. The <li> for search should include a class of rhd-nav-search and the contents should look like:

```
<div class="rhd-search-wrapper">
    <form class="search-bar" accept-charset="utf-8" action="/search/" method="get" role="search" type="submit">
        <input class="user-search" type="text" placeholder="Search" id="search_list_text" name="t" maxlength="128" required="">
        <button type="submit" id="search-button"><i class="far fa-search"></i></button>
    </form>
</div>
<button class="rhd-search-toggle rhd-search-toggle-open"><span class="open-icon"><i class="far fa-search"></i></span></button>
<button class="rhd-search-toggle rhd-search-toggle-close"><span class="close-icon"><i class="far fa-times"></i></span></button>
```

Right now the list element looks like:

```
<li class="pf-c-nav__item pf-u-mr-0">
    <a href="#" class="pf-u-display-inline-block pf-u-pt-md pf-u-pb-sm"><i class="fas fa-search"></i></a>
</li>
```

PR #3191, opened 2019-10-14
[NO ISSUE] Load PFElement in the head

jordanpagewhite:
This loads the PFElement JS file in the head and no longer loads the
onebox JS file explicitly, as that is loaded via the Search JS file.

### JIRA Issue Link

n/a

### Verification Process

- `js/pfelement.umd.min.js` is loaded in the head.
- There are no JS errors in the console on the /search page, or at least none specifically related to our web component implementation.


robpblake:
rebuild this please

Issue #3203, opened 2019-10-15
Fix title on templates for download page node view modes


fearlsgroove:
Product Download Page
Product Getting Started Page

jordanpagewhite:
This was closed by #3228

PR #3228, opened 2019-10-15
[SURGE-3203] Add templates Product download and getting started

jordanpagewhite:
This adds Twig templates for the Product /download and /getting-started
subpages. This includes the changes required to remove the page title,
as we do not want that in our assembly-based content strategy.

### JIRA Issue Link

#3203 

### Verification Process

Go here: https://localhost/products/codeready-workspaces/download 
and here: https://localhost/products/codeready-workspaces/getting-started

You should **not** see a page title at the top of the page


jordanpagewhite:
I demo'd this on the Hangouts call. We demo'd and approved both the 'no page title' issue for /overview /getting-started and /download subpages and the changes that Alistair made earlier today in the FE repo to the Product Download Hero.

PR #3259, opened 2019-10-16
[NO ISSUE] Display Code Snippet Assemblies

jordanpagewhite:
A regression was accidentally introduced here:
https://github.com/redhat-developer/developers.redhat.com/commit/7749cc4011c411c5f31f2c46b55b4e5436101d53#diff-304ef1e9f8b0e28569a944480f0512da
that removed a necessary variable for rendering the code snippet
assembly. I have added back the preprocessing of that variable, while
still leaving out the attachment of the highlight-js library, as that is
included globally to accommodate all `<code><pre>` elements. I used UX's
Marvel design to determine the 8px of padding.

### JIRA Issue Link

n/a

### Verification Process

Go here: `/coderland/serverless/building-a-serverless-service/`

You should see this:

Screen Shot 2019-10-16 at 12 54 48 PM

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/themes/custom/rhdp2/css/rhdp2.css
+++ b/_docker/drupal/drupal-filesystem/web/themes/custom/rhdp2/css/rhdp2.css
@@ -638,5 +638,5 @@ a.pf-c-button:hover {
 
 /* Code Snippet */
 .assembly.assembly-type-code_snippet .hljs {
-  padding; 8px;
+  padding: 8px;
 }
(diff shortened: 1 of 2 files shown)


robpblake:
retest this please

staceymosier:
@jordanpagewhite 

I'm not certain what the class hljs means, but since that's where the background color is defined, I added width: 100%. 

(I only changed it here from gray to yellow for this screenshot so that it is apparent. The gray/white contrast isn't strong on my monitor.)

I don't see a component named "Code Snippet" in [PatternFly](https://pf4.patternfly.org/) to know if this is the intended style.

assembly-code-snippet

PR #3261, opened 2019-10-16
[SURGE-347] Node Reference Cards spacing

jordanpagewhite:
This resolves an issue where Node Reference cards, in static content
bands, had too much spacing between cards in the card grid.

### JIRA Issue Link

https://github.com/redhat-developer/rhd-frontend/issues/347

### Verification Process

Go here: https://localhost/topics/kubernetes/

You should see this:

Screen Shot 2019-10-16 at 1 02 50 PM

and

Screen Shot 2019-10-16 at 1 05 02 PM

NOTE: I created a follow-up issue for that right margin on the cards on mobile here: https://github.com/redhat-developer/rhd-frontend/issues/384


gdoyle1:
@jordanpagewhite Alistair is working on a related card fix... since our design requirements changed a bit for curated content (based on content editor feedback) I think 1) whatever you write will get overwritten with his fixes and 2) we want something else other than the solution I proposed in the beginning.

PR #3265, opened 2019-10-16
[NO ISSUE] Add back marketing scripts and Keycloak toggle

jordanpagewhite:
This adds back some necessary preprocessing for marketing script
variables in html.html.twig, and the commit also adds back a toggle for
Keycloak to provide Keycloak stage on non-Prod environments.

### JIRA Issue Link

n/a

### Verification Process

Testing steps TBD. I am not certain if we can test these marketing integrations on PR environments.

@robpblake For Keycloak, we create the code in the hook below `function rhdp2_library_info_alter(array &$libraries, $extension)` during the Managed Platform epic, but I am confused now because I am certain that we were loading Keycloak stage on local environments without this hook. Are we doing anything in the build layer to toggle that?

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/themes/custom/rhdp2/rhdp2.theme
+++ b/_docker/drupal/drupal-filesystem/web/themes/custom/rhdp2/rhdp2.theme
@@ -718,3 +718,174 @@ function rhdp2_form_alter(&$form, \Drupal\Core\Form\FormStateInterface $form_sta
    $form['#attached']['library'][] = 'rhdp2/drupal-user-login-form';
  }
 }
+
+/**
+ * Implements hook_preprocess_html().
+ */
+function rhdp2_preprocess_html(array &$variables) {
+  $current_path = \Drupal::service('path.current')->getPath();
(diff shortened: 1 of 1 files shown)


robpblake:
@jordanpagewhite There are two parts to the Keycloak integration:

1. The server you use to perform the authentication
2. The location where you load `keycloak.js` from.

For various reasons, it's not practical for us to use prod infra in the PR environments. For example logins using prod Keycloak cannot cope with the fact that we redirect to multiple endpoints using a single client identifier e.g. our PR client redirects to https://developer-preview-1,  https://developer-preview-2 and so on. 

This in turn impacts our ability to test downloads in PR environments. So the best compromise we can have is to use the stage Keycloak for authentication (to support our wildcard redirecting) but continue to use the prod `keycloak.js` to ensure our front-end works as expected.

In stage we _should_ be overriding the library definition to use the staging Keycloak JS, but we're not, so that is an omission which we can address elsewhere, but not in this PR.

On another note I'm 99% sure this is why the mobile tests were failing now. I'll merge this and then try another PR with the mobile tests re-enabled.

cheers,

Rob

PR #3287, opened 2019-10-17
[MWES-3480] Remove Phinx from Drupal boot process

robpblake:
Two changes in this commit.

Firstly I'm removing the execution of Phinx from the Drupal boot process and the load of the migrations into the built Drupal container. I cannot remove the creation and mount of phinx.yml in this commit as I will need to update the Openshift templates for our deployment first. Chicken and egg situation.

Secondly I'm removing the setup_local_drupal.sh and remove_local_drupal.sh scripts. These are now out-of-date to the point of being dangerous for local development, so no one should be using them.

### Verification Process

* The build should go green


robpblake:
rebuild this please

robpblake:
rebuild this please

robpblake:
rebuild this please

PR #3303, opened 2019-10-24
DEVELOPER-5943 - Update to latest Sentry

KamiQuasi:
### JIRA Issue Link
* https://issues.jboss.org/browse/DEVELOPER-5943

### Verification Process
This combines updating our JS error tracking via Sentry.IO.


robpblake:
retest this please

_docker/drupal/drupal-filesystem/web/themes/custom/rhdp/rhd-frontend/src/scripts/@rhd/js/drupal-namespace.js
```diff
@@ -133,7 +133,7 @@ app.products = {
 };
 
 app.products.downloads = {
-    "cdk" : {"windowsUrl" : "/download-manager/file/cdk-3.9.0-1-minishift-windows-amd64.exe", "macUrl" : "/download-manager/file/cdk-3.9.0-1-minishift-darwin-amd64", "rhelUrl" : "/download-manager/file/cdk-3.9.0-1-minishift-linux-amd64"}
+    "cdk" : {"windowsUrl" : "/download-manager/file/cdk-3.10.0-1-minishift-windows-amd64.exe", "macUrl" : "/download-manager/file/cdk-3.10.0-1-minishift-darwin-amd64", "rhelUrl" : "/download-manager/file/cdk-3.10.0-1-minishift-linux-amd64"}
```
robpblake: @KamiQuasi Is there any reason we cannot read this directly from Download Manager?

I'll leave @jordanpagewhite to comment, but I think this is also going to need to be updated in `rhd-frontend` as well as this file is not used by the `rhdp2` theme.

jordanpagewhite: @KamiQuasi I am also wondering why we can't just fetch this data from Download Manager. It seems very risky to have this hidden away in a file like this. Rob is right that this will only change this value in the `rhdp` theme, which is only loaded for old, paragraphs-based Product pages. That being said, the `cdk` Product page _is_ an old, paragraphs-based Product. If a content editor were to update/change this Product to the new, assembly-based approach though, this data would not be available.

It seems to me that this would best be resolved in Download Manager unless there is some compelling reason that we (developers.redhat.com) have some unique use-case that deviates from other consumers of Download Manager.

_docker/drupal/drupal-filesystem/web/themes/custom/rhdp/templates/system/html.html.twig
```diff
@@ -247,8 +247,15 @@ window.NREUM||(NREUM={}),__nr_require=function(t,n,e){function r(e){if(!n[e]){va
...
-    <script>Raven.config('{{ rhd_sentry_code }}', { environment: '{{rhd_environment}}', whitelistUrls: [/redhat\.com/] }).install()</script>
+    <script src="https://polyfill.io/v3/polyfill.min.js?features=Promise%2CObject.assign%2CString.prototype.includes%2CNumber.isNaN"></script>
+    <script src="https://browser.sentry-cdn.com/5.7.1/bundle.min.js" integrity="sha384-pHnCJ3bbzJ+LzI19wAdyBvGAjJ5vNAC59uxeRpw5WpgkSlZUQstEADxkSbBnI95q" crossorigin="anonymous"></script>
+    <script>
+    Sentry.init({ 
+      dsn: '{{ rhd_sentry_code }}',
+      environment: '{{rhd_environment}}',
+      whitelistUrls: [ /developers\.redhat\.com/ ]
+    });
+    </script>
```
jordanpagewhite: I've created a ticket here: https://github.com/redhat-developer/developers.redhat.com/issues/3307 to move all of these scripts that are defined in the `html.html.twig` files into Drupal libraries. I think that would make it clearer, with respect to code organization, and hopefully we can optimize the way that we contextually load some of these scripts and styles down the road.

jordanpagewhite: That is a net-new feature in my opinion though, and I am happy with these changes. It does not particularly matter to whether the `rhd_sentry_script` variable is used or the URL is manually declared here. It looks like the variable isn't even set by the build processes or dynamic, based upon the environment, so I do not see much of a difference personally.

robpblake: @jordanpagewhite The value of the `rhd_sentry_script` variable is set in `rhdp.theme` and `rhdp2.theme` and is read from `rhd.settings.php`

- https://github.com/redhat-developer/developers.redhat.com/blob/master/_docker/drupal/drupal-filesystem/web/themes/custom/rhdp/rhdp.theme#L154-L156

and

- https://github.com/redhat-developer/developers.redhat.com/blob/master/_docker/drupal/drupal-filesystem/web/themes/custom/rhdp2/rhdp2.theme#L742-L744

So if we want to hard-code it into the template (which is fine if it never needs to vary between environments) then I'd like to see us clear up the properties from `rhd.settings.php`

robpblake:
@KamiQuasi Where we at with this one?

cheers,

Rob

robpblake:
retest this please

robpblake:
retest this please

KamiQuasi:
@jordanpagewhite and @joeschram if RHD isn't going to keep paying for Sentry tracking this PR can be closed, and the Sentry-specific code removed.

jordanpagewhite:
I need to review this today. I wasn't sure if we were keeping our Sentry.io subscription, but we are.

jordanpagewhite:
I am going to review this locally since I need to toggle config variable(s) that

alistairmc:
@jordanpagewhite this seems to of been hanging around for a while did you get it tested? is it still valid?

jordanpagewhite:
It does not seem that this work will be prioritized any time in the foreseeable future, and we need extra resources in Openshift, so I'm going to close this

Issue #3304, opened 2019-10-24
Local development experience runs 3 theme builds by default


robpblake:
To simply start Drupal locally, we are currently running npm processes for:

* `rhd-frontend` in `rhdp` theme
* `rhd-frontend` in `rhdp2` theme twice

This is adding significant overhead to the local development experience, particularly if you don't need to make any changes to the theme.

Related to: #3065

Issue #3065, opened 2019-09-10
Proposal for integration of RHD Frontend Assets in Drupal


robpblake:
I wanted to document my proposal with Drupal specific implementation details as I don't think I'm conveying it very well. Hopefully this better explains how I see things integrating.

**Goals of the integration:**

- Production of the `rhd-frontend` assets to be consumed by Drupal is performed by a CI process on the `rhd-frontend` project
- The Drupal container has zero need to run the `rhd-frontend` build process e.g. `npm install`
- The `rhd-frontend` project is not included in the Drupal container at build time or run time.
- `rhd-frontend` does not need to be included as a Git sub-module in the `developers.redhat.com` repository
- Local development on the `rhd-frontend` assets is simple and does not impact the local Drupal dev experience if you don't need to work with `rhd-frontend` 
- Support easily switching the version of the `rhd-frontend` assets used in different environments
- Be as idiomatic to Drupal as possible when integrating this stuff
- Follow closely our existing patterns for integrating with external dependencies e.g. Download Manager

**rhd-frontend assets are configuration** 

This proposal works on the basis that where Drupal loads the `rhd-frontend` assets from is environment configuration in exactly the same way that we configure the Download Manager URL or Keycloak URL in each environment. Drupal will expect to find the `rhd-frontend` assets at the URL you configure it with. How those assets get there is not the concern of Drupal or the Drupal CI build process. Drupal simply expects that the assets will be where you tell it they will be.

Most likely, we will configure the CI process for the `rhd-frontend` to publish assets to a location that an be easily consumed by Drupal e.g https://prod.redhatstatic.com/developer`. Additionally, the assets from `rhd-frontend` will be published in such a way that versioning is explicit in the URL. My proposal is that the URL of the assets will contain the GIT SHA of the HEAD of the `rhd-frontend` repository at the time the assets were published. So if the GIT SHA was `12345` at the time the `rhd-frontend` assets were published, you would expect to find them at:

- https://prod.redhatstatic.com/developer/rhd-frontend/12345/ e.g.
   - https://prod.redhatstatic.com/developer/rhd-frontend/12345/css/rhd.min.css
   - https://prod.redhatstatic.com/developer/rhd-frontend/12345/js/rhd.min.js
   - https://prod.redhatstatic.com/developer/rhd-frontend/12345/favicons

**Configuring Drupal with the `rhd-frontend` asset URL**

[...]

PR #3305, opened 2019-10-24
[#3304] Remove multiple builds of rhd-frontend theme

robpblake:
This commit does a couple of things.

Firstly it removes any reference to the fontawesome stuff from the Drupal build. We are now hosting those assets from a CDN.

Secondly it only calls `npm run-script build` to create the new rhd-frontend assets. We were previously calling `npm run-script styles` and then `npm run-script build`, however the build script additionally calls styles, so this was a duplication

### Verification Process

* The build should go green
* The site should look as expected

### DIFF
--- a/_docker/drupal/Dockerfile.mp
+++ b/_docker/drupal/Dockerfile.mp
@@ -36,10 +36,7 @@ RUN cd /var/www/drupal/web/themes/custom/rhdp/rhd-frontend \
 ARG FONTAWESOME_LICENCE
 RUN cd /var/www/drupal/web/themes/custom/rhdp2/rhd-frontend \
     && npm config set "registry" https://repository.engineering.redhat.com/nexus/repository/registry.npmjs.org/ \
-    && npm config set "@fortawesome:registry" https://npm.fontawesome.com/ \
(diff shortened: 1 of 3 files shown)


robpblake:
> This looks good to me. Thanks Rob. This is perhaps added scope to what you're looking to do with this particular PR, but do we need to execute `cd "${DIR}" && ./build-theme.sh` on line 42 of `dev/run-drupal.sh`? If I am following the chain of build commands properly, I believe the theme will already be built by Dockerfile.mp in the `drupal` job of `docker-compose.yml`. Is that right?

@jordanpagewhite You are correct, but the only problem is that we then overwrite the theme directory through a volume definition in the `docker-compose.yml` for local dev (this was done to allow you guys to re-build the theme locally whilst working on the surge). If we don't add in the additional re-build of the theme, then Drupal will start completely un-styled. 

There are lots of things that we can do to fix the local dev experience with regards to using the rhd-frontend component, but it needs discussion and agreement e.g #3065 , https://github.com/redhat-developer/rhd-frontend/issues/395 and https://github.com/redhat-developer/rhd-frontend/issues/203 and as you point out, I'd like to keep this PR focussed on the one issue identified.

cheers,

Rob

PR #3306, opened 2019-10-24
[MWES-3660] First steps in migration to merged SSO servers

robpblake:
This is the first step in migrating to the new SSO servers hosted by IT. In this commit I'm simply exposing the realm and client_id as additional configuration parameters that will facilitate an easier migration because things no longer need to be hard coded. This commit does not alter the current approach to login or the SSO servers in use.

### JIRA Issue Link
* https://issues.jboss.org/browse/MWES-3660

### Verification Process

* The build should go green
* Open a javascript console and ensure that the following are set:
  * drupalSettings.rhd.keycloak.client_id = 'web'
  * drupalSettings.rhd.keycloak.realm = 'rhd'
  * app.ssoConfig.client_id = 'web'
  * app.ssoConfig.realm = rhd'

The last two will only be available on the old style product pages for now until I make an additional update to the rhd-frontend project.


robpblake:
rebuild this please

jordanpagewhite:
@robpblake The only thing that I will point out, which you might already be aware of, is that the following variables will not be available (for example from your Verification Steps) other than on old, paragraphs-based Product pages:

app.ssoConfig.client_id = 'web'
app.ssoConfig.realm = rhd'

These are being set in the `rhd` theme only, which is only served on the old, paragraphs-based Product pages. If we want these variables to be available everywhere, you will have to include a PR to the rhd-frontend repo and update the git submodule to reference the commit with those changes.

robpblake:
> @robpblake The only thing that I will point out, which you might already be aware of, is that the following variables will not be available (for example from your Verification Steps) other than on old, paragraphs-based Product pages:
> 
> app.ssoConfig.client_id = 'web'
> app.ssoConfig.realm = rhd'
> 
> These are being set in the `rhd` theme only, which is only served on the old, paragraphs-based Product pages. If we want these variables to be available everywhere, you will have to include a PR to the rhd-frontend repo and update the git submodule to reference the commit with those changes.

@jordanpagewhite Yeah. So I'm gonna have to co-ordinate updates to both of the repos. My thought was first and foremost just to get the properties exposed and then I'll tackle that one. I'm sending out an email today with plan of action for migration to merged SSOs.

cheers,

Rob

PR #3308, opened 2019-10-25
Searching from the mobile menu does not work #3196

### LINKED ISSUE #3196: Searching from the mobile menu does not work
If I open the mobile menu, enter a search term and then click on the magnifying glass, nothing happens. Additionally pressing enter does not seem to search either.

alistairmc:
### JIRA Issue Link
Fixes #3196 

### Verification Process
Test that magnify glass on desktop search or the return key submit the search in the nav menu on desktop or mobile


jordanpagewhite:
I just tested this with the following steps and everything worked as intended:

Go here: https://developer-preview-3308.ext.us-west.dc.preprod.paas.redhat.com

Do this:

- Open the mobile inspector (iPhone 6/7/8) in Chrome
- Click the Menu hamburger icon toggle button
- Enter a search query in the search input in the mobile menu
- Click the Search icon

You should see this:

- You should be on the /search page and you should see search results for your query like https://developer-preview-3308.ext.us-west.dc.preprod.paas.redhat.com/search/?t=istio

Do this:

The same exact steps and results  as above but press the Enter key to search instead of clicking the search icon.

Issue #3119, opened 2019-09-25
Remove topic page display NOT view_topics


jordanpagewhite:
We have a `view_topics` that display here: `/topics` and a deprecated `topic` with an unpublished `Page` display. We can remove/delete that `Page` display, but we should keep the `Feed` display, as the RSS Feed is used.

rogeralloymagnetic:
PR: https://github.com/redhat-developer/developers.redhat.com/pull/3353

PR #3353, opened 2019-11-06
[DEVELOPER 3119] Remove Page display from topic view

ghost:
### JIRA Issue Link
* https://github.com/redhat-developer/developers.redhat.com/issues/3119

### Verification Process
1. Login to Drupal
2. Go to Views sections
3. Click `Edit` next to the `topic` view
4. Confirm there is only a `Feed` display and there is not a `Page` display


jordanpagewhite (approved):
I just reviewed this and confirmed that the Page display on the `topic` View no longer exists and the Feed display still exists at the same path https://developer-preview-3353.ext.us-west.dc.preprod.paas.redhat.com/topic/feed

PR #3389, opened 2019-11-19
Fix left align issue and float on assembly-type-static_content_band

### LINKED ISSUE #3283: Curated Content Assembly: Left align by default
Estimate: 2

This assembly should be left align by default. Currently it is centerer:
Screen Shot 2019-10-16 at 9 48 27 PM

And, when you apply the left aligned visual style, this madness happens:
Screen Shot 2019-10-16 at 9 47 29 PM

That being said, there should be no left align visual style. That should be replaced with center.

### LINKED ISSUE #3294: Curated content is centered when adding 3 and 4 pieces of content
Estimate: 2

This assembly needs to be left aligned by default. Right now it is centered.

Screen Shot 2019-10-18 at 8 23 13 AM

Ex on /coderland/reactica page. However, that assembly is working right on the /coderland/serverless page. 

There are the two differences between the two pages above:
- /coderland/serverless page: Using the 4-up visual style. This should not affect the alignment and also no longer applies so should remove this visual style. 
- The "node reference" content type centers the content as well, while "static item" displays correctly. /coderland/serverless page uses static items whiles /coderland/reactica page uses node reference.

alistairmc:
### JIRA Issue Link
Fixes #3294 
Fixes #3283 

### Verification Process
go to /coderland/reactive/ and check the 'watch the videos' section behaviour in various screen sizes.

Also, check /topics/devops/ 'Getting started with DevOps' section

Added a centered class to the component, in the admin go to either of the above pages and remove the 4 up class and add the centered option

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/config/sync/assembly.assembly_type.static_content_band.yml
+++ b/_docker/drupal/drupal-filesystem/web/config/sync/assembly.assembly_type.static_content_band.yml
@@ -5,5 +5,5 @@ dependencies: {  }
 id: static_content_band
 label: 'Curated Content'
 description: 'An assembly that displays a curated list of wordpress posts and/or drupal nodes.'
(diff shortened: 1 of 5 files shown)


alistairmc:
@gdoyle1 can you take another look at this I have added the centered calls for the admin, if you log in you can test it.

I am not 100% happy with the solution but its close, there seems to be quite a bit of variation in the implementation of cards in grids and its something we are going to need to look into moving forward

robpblake:
retest this please

PR #3446, opened 2019-12-18
[RHDX-82] Rework cards for static band

alistairmc:
### JIRA Issue Link
https://projects.engineering.redhat.com/browse/RHDX-82

### Verification Process
check card alignment on homepage two column assembly

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/config/sync/assembly.assembly_type.static_content_band.yml
+++ b/_docker/drupal/drupal-filesystem/web/config/sync/assembly.assembly_type.static_content_band.yml
@@ -5,5 +5,5 @@ dependencies: {  }
 id: static_content_band
 label: 'Curated Content'
 description: 'An assembly that displays a curated list of wordpress posts and/or drupal nodes.'
-visual_styles: "no-padding-top|No padding top|Remove top padding from this assembly\r\nno-padding-bottom|No padding-bottom|Remove bottom padding from this assembly\r\ndark|Dark|Dark background with white text\r\ncentered|Center|Center aligned content\r\nfour-up|4-up|When displaying items in a row, show 4 instead of 3"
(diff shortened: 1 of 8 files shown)


alistairmc:
@sjcox-rh @gdoyle1 need some feedback on this, I have two main questions.

1. as the screen reduces the cards reduce in size do you think I need to break them to two-column sooner ( IE at a bigger breakpoint ).

2. I noticed on live the images were cropping text off on some screen sizes, I have implemented the blurred background images in this case. Do you want this as the standard option with a class to be able to enable the old cropped style?

sjcox-rh:
@alistairmc answering your questions below:

1. Just reviewed and tested out the various breakpoints and I think you're right, we should break at two columns sooner. Right now it seems that it breaks to single columns starting at medium but perhaps we should push it to the small breakpoint. Having cards that are 600+px width is too wide.

I'm not sold on it and need to see an example because I'm concerned that the cards might be too small right before the small breakpoint of 480. But it may work!

2. I would chat with @staceymosier about this. I know she's working on a spike to standardizing how images are cropped throughout cards. This may be an option we want to consider but not sure we should merge this class at the time being until we all decide on an approach.

3. Is there a way to retain the fixed height across cards in a static band without the blurred bg image for the time being?

alistairmc:
@sjcox-rh I can have the height fixed but images will crop and you could chop text this can be mitigated bu using multiple images for different breakpoints ( The work Stacy is looking at ). I will default to the cropped image style and fix the height. Also, I will look at moving the breakpoints as you suggested.

alistairmc:
@gdoyle1 @sjcox-rh I have made the suggestions you mentioned and have experimented with the breakpoints for the cards in the two col hero, let me know what you think.

robpblake:
retest this please

PR #3451, opened 2019-12-19
[RHDX-118] Display Image field, hide Media reference

jordanpagewhite:
This config export will now make it so that the Paragraph with Image
paragraph type displays its Image field on both the paragraph edit form
and when viewing the paragraph. It also removes the Media reference
field from display on the form and when viewing the Paragraph. There is
a regression with the implementation of the Media reference field, from
what I can tell dating back to June 13th, 2017, that makes the field
widget display on the paragraph edit form but entirely fails to render
it when viewing the paragraph due to our completely custom rendering
process of the old, Paragraphs-based Product pages.

### JIRA Issue Link
* https://projects.engineering.redhat.com/browse/RHDX-118

### Verification Process

Go here: https://developer-preview-3451.ext.us-west.dc.preprod.paas.redhat.com/user/login

Do this:

* Log into an account with only the Content Creator role (create a test account if necessary)

Go here: https://localhost/node/37375/edit

Do this:

* Click on the Fields tab
* Scroll down to the Individual Sub Pages section
* Click Edit for the Overview product page (paragraph)
* Scroll down a little bit and find the PRODUCT FEATURE IMAGE AND TEXT field and the Media Reference field within that paragraph reference (you might need to click Media Reference to toggle the visibility)
* Click the Add Media button
* Click the Upload tab in the modal that loads
* Click 'choose file' and select any image from your local machine
* Enter the required Name field for the Media entity and click the blue Place button
* Verify that your image now displays as a thumbnail in the node edit form (upon persisting that modal form)
* Scroll back up to the top of the page and click the Settings tab
* Enter some revision message
* Click the blue Save button in the top-right corner to persist the changes to the RHAMT product page
* Verify that your image now displays as one of 3 images on this page: https://localhost/products/rhamt/overview


robpblake:
retest this please

staceymosier:
@jordanpagewhite 

For the Paragraph with Image,  the Image displays, but the Media reference doesn't.
I removed the Media Reference value and tried to save, but I had form validation message.

> Media Reference field is required

1. There is no indication in the Form that the field is required.
2. The media does not display on the page.  Why is the field required, but not in use?

See the Command Line Tooling paragraph here
https://developer-preview-3451.ext.us-west.dc.preprod.paas.redhat.com/products/rhamt/overview

jordanpagewhite:
Thanks. Something in config isn't matching up with my local. I will take a look and resolve this after our morning scrum call

staceymosier:
I am able to save the product 'Paragraph with Image' field without attaching an image. This results in a broken page with the error

`TypeError: Argument 1 passed to Drupal\Core\Entity\EntityViewBuilder::view() must implement interface Drupal\Core\Entity\EntityInterface, null given, called in /var/www/drupal/web/modules/custom/rhd_common/rhd_common.product_related.inc`

PR #3465, opened 2020-01-13
[No Issue] Further tighten access to key Drupal resources

robpblake:
This small change to the Drupal Apache configuration further tightens access to key Drupal pages. 

### JIRA Issue Link
* N/A

### Verification Process

* The build should go green
* Verify that you can log in as expected using the SSO login
* Log in to the Drupal container and ensure `drush cron` works as expected
* Ensure that the following pages result in a 403 Forbidden response:
  * install.php
  * cron.php
  * authorize.php
  * upgrade.php
  * xmlrpc.php


jordanpagewhite:
I was just able to authenticate via SSO. I have multiple accounts on Stage and made a user error. Thanks.

PR #3471, opened 2020-01-15
[RHDX-148] Fix background position on CTA component

alistairmc:
### JIRA Issue Link
https://projects.engineering.redhat.com/browse/RHDX-148

### Verification Process
go to /products/openjdk/overview the cta background should run to the edge of the component

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/themes/custom/rhdp2/rhd-frontend/src/styles/rhd-theme/components/_cta.scss
+++ b/_docker/drupal/drupal-filesystem/web/themes/custom/rhdp2/rhd-frontend/src/styles/rhd-theme/components/_cta.scss
@@ -1,35 +1,37 @@
 @import "../mixins";
-
-.assembly-type-call_to_action {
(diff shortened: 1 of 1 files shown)


alistairmc:
rebuild this please

jordanpagewhite:
That is a great point @staceymosier Thanks. It would be great to combine these files if possible since they both seem to target the same assembly type/component

alistairmc:
@staceymosier @jordanpagewhite 
 I don't think the 2018 one is loaded anymore

jordanpagewhite:
@alistairmc Yep, sorry about that. You are right. I had to read through the various `@import` statements, but it looks like that file isn't processed into a dist asset at all. I can create a separate issue to perhaps clean up some of the legacy SCSS that is only causing clutter in the `rhdp2` theme then, but that will not be in the scope of this issue/PR. Thanks Alistair.

jordanpagewhite:
@staceymosier Can you give this another review, considering Alistair's comment above, when you have a chance? Thanks.

PR #3495, opened 2020-02-06
[RHDX-248] Update Layout: 2 Column assembly type display

jordanpagewhite:
This changes the way that Layout: 2 Column assembly type will display
content. We want to use this assembly type to support any kind of
assembly-based, 2 column layouts, not just a left sidebar scenario. To
do so, I have created 2 visual styles: Left sidebar and right sidebar.
The default behavior, now, will be an evenly split 6:6 layout.

### JIRA Issue Link
* https://projects.engineering.redhat.com/browse/RHDX-248

Relates to: https://projects.engineering.redhat.com/browse/RHDX-179

#### Default - no visual style
localhost_ (4)

#### Right sidebar visual style
localhost_assembly_31000_view

#### Left sidebar visual style
localhost_ (5)

### Important note

If/when merged, this will require a content update (update the visual style to Left Sidebar) of the 1 referenced Layout: Two Column assembly on Prod (the homepage). We could write a migration for that I suppose, but since there is only 1, it seems like a better use of time to just update the content manually.

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/config/sync/assembly.assembly_type.layout_two_col_left_sidebar.yml
+++ b/_docker/drupal/drupal-filesystem/web/config/sync/assembly.assembly_type.layout_two_col_left_sidebar.yml
@@ -3,7 +3,7 @@ langcode: en
 status: true
 dependencies: {  }
 id: layout_two_col_left_sidebar
-label: 'Layout: Two Column with Left Sidebar'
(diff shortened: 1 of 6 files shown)


jordanpagewhite:
@sjcox-rh Yes, this will be used to address use cases like the Eloqua form issue that Gina experienced

PR #3496, opened 2020-02-06
[RHDX-221] Add webinar countdown component

staceymosier:
### JIRA Issue Link
[RHDX-221](https://projects.engineering.redhat.com/browse/RHDX-221)

### Verification Process

View sample webinar page here: https://developer-preview-3496.ext.us-west.dc.preprod.paas.redhat.com/node/214685/

**Scenario 1** 
Create a new page with a Call to Action assembly, or edit a Call to Action assembly.
Provide a value for CTA Timer and CTA Timer Label for a few minutes in the future.
Use the Left visual styles so that content flushes left. Use the Dark visual style and provide a dark background image.
Verify the Countdown is accurate for your timezone.
At 0, it should be hidden and the CTA button should appear.

**Scenario 2**
Verify the Call to Action assemblies without Countdown values still work as expected.

**Scenario 3**
Edit the CTA from Scenario 1 and remove the Left visual style.
Check how it renders as centered.  Check how it renders a Light visual style and light background image.

**Note:**  This assembly is using pf-l-grid 8 columns to keep the text from overflowing the lightbulb.
https://developer-preview-3496.ext.us-west.dc.preprod.paas.redhat.com/node/214685/


staceymosier:
rebuild this please

jordanpagewhite:
@gdoyle1 and @staceymosier Re: #4 in Gina's comment above,

I think that we should explore what use cases exist that could require this countdown component to be placed in a Rich Text assembly. I assume that these scenarios are motivated by layout possibilities? Are there additional concerns? We should definitely discuss these. I would like to identify and fully define the problem before we discuss potential solutions.

jordanpagewhite:
Also, I reviewed the current state of the code and PR environment, and the changes look good to me, but I will wait to approve this until we have resolved Gina's comment 4 above.

staceymosier:
@gdoyle1  I pushed up a change to fix the center display on the default assembly (i.e. without the left visual style).  I also checked it on mobile;  the max width it will grow to be is 400px.

I did not add the countdown to the rich text assembly.  The call to action assembly has the 'call to action' link field that can be targeted specifically to hide/show.  The editor has the option of putting link/button markup in the body of the cta field, however it will appear before the countdown.

When the PR rebuilds, I'll put the sample content back and request another review.

gdoyle1:
> @gdoyle1 and @staceymosier Re: #4 in Gina's comment above,
> 
> I think that we should explore what use cases exist that could require this countdown component to be placed in a Rich Text assembly. I assume that these scenarios are motivated by layout possibilities? Are there additional concerns? We should definitely discuss these. I would like to identify and fully define the problem before we discuss potential solutions.

@sjcox-rh I think we originally thought we'd want to place it into a rich text assembly since that provides the most flexibility as a content creator BUT thinking about the use cases where the countdown component will be used, it will just be for these webinar pages. With that said, everything that was recreated in the sample page using the Call to action assembly was exactly how it is in Drupal. So, after I make sure the center visual style works, I think we can merge this from a UX standpoint!

SJ if you could verify that that is correct too that'd be great ^

sjcox-rh:
@gdoyle1 you are correct. We originally requested that we have this feature in all 3 rich text components BUT Gina made a good point that we could still create the same experience with the call to action assembly so I'm okay with merging it as it stands. 

If we get to a point in the feature where we would need this component in more assemblies, then we can just revisit this.

@staceymosier I did have a tiny question. Does this ticket as address the behavior what happens when the countdown ends?

Will it turn into the CTA?

staceymosier:
@sjcox-rh    While the countdown is appearing, at 0000, it will hide and the CTA button from the assembly will appear.    It checks the time against Eastern NewYork time, so any time value entered into the field must be from that range.  If you refresh the page and the timer is already out, it will show the CTA button.

sjcox-rh:
@staceymosier thats perfect!  I can't remember, but it there a note next to the field/field label where you input the time that clearly states that it's based off Eastern Standard Time?

Other than that, this is all correct!

jordanpagewhite:
Thank you @sjcox-rh . That is a good point. How does this sound to you both? cc @staceymosier 

#### Display CTA Timer

Provide a datetime (EST / GMT-05:00) if you would like to hide a CTA button and display a countdown/timer ex. preceding an event like a webinar. Once this datetime has been reached, the countdown/timer will be hidden and the CTA will display.

#### Display CTA Timer Label

This text will display over the timer/countdown component ex. "Event starting in...", "Event coming soon", "Webinar available in...", etc.

staceymosier:
I added the descriptions to the 2 fields.

I'm working on cleaning up some visual styles for Dark so that `.pf-l-grid` classes can be used in the call to action field;  I will remove DNM when it's ready for another review.

gdoyle1:
@staceymosier for some reason the time I'm entering isn't showing up correctly - I entered 2:09pm when it was 2:08pm in real time and it said i had 4 hours to wait?

_docker/drupal/drupal-filesystem/web/modules/custom/rhd_assemblies/rhd_assemblies.module
```diff
@@ -128,15 +128,28 @@ function rhd_assemblies_entity_presave(EntityInterface $entity) {
...
+  $type = $assembly->bundle();
 
   if ($parent) {
     // Make assembly ID match pattern used for on page navigation.
     $id = $variables['elements']['#assembly']->id();
     $css_id = Html::cleanCssIdentifier('assembly-' . $parent['field_name'] . '-' . $id);
     $variables['attributes']->offsetSet('id', $css_id);
   }
+
+  // Call to Action Countdown Component
```
jordanpagewhite: PHPCS: I think we need a period at the end of this comment/line.

staceymosier:
rebuild this please

robpblake:
retest this please

PR #3518, opened 2020-03-02
[RHDX-286] Split out patternfly imports / update base component styles

alistairmc:
### JIRA Issue Link
https://projects.engineering.redhat.com/browse/RHDX-286

### Verification Process
Check site still functions as before, the site appearance should be unchanged.


jordanpagewhite:
I clicked around the most popular pages and page types on the site, and the site functionality and appearance seems unchanged. Also, the typography styles, noted in the JIRA ticket, are not  being applied as they are on Prod

alistairmc:
@staceymosier I have moved the class pf-c-content to the assembly/component level seems to make more sense living there.

PR #3524, opened 2020-03-09
[RHDX-333] Make DownloadManagerApi use EntityTypeManagerInterface

jordanpagewhite:
This is a change to properly inject EntityTypeManagerInterface as a
dependency, instead of EntityTypeManager. I discovered this issue while
debugging some performance issues when I tried to enable and use the
webprofiler module. It will result in a fatal error because we are not
properly injecting the interface.

### JIRA Issue Link
* https://projects.engineering.redhat.com/browse/RHDX-333

### Related work

* I will be creating a merge request to Compose soon because it also includes 1 instance of this same issue. I will track that in the JIRA issue.

### Verification Process

* Everything works exactly the same as it did before
* Unfortunately, you will not be able to test this next step until we make a change to the Compose (contrib) module, and we cannot use the latest Compose branch until we are on Drupal core 8.8.x. This is the scenario, though, that originally uncovered this issue.
** You can check out this branch and enable and use the `webprofiler` Devel module


jordanpagewhite:
rebuild this please

robcoffman:
rebuild this please

robpblake:
rebuild this please

robpblake:
rebuild this please

PR #3536, opened 2020-03-31
[RHDX-339] Resolve php warning for ddo_taxonomies

staceymosier:
### JIRA Issue Link
[RHDX-339](https://projects.engineering.redhat.com/browse/RHDX-399)

### Verification Process
View any product path that does not exist so that you see a 404.
View the dblogs and verify there are no PHP warnings with the message `Notice: Undefined variable: ddo_taxonomies in redhat_www_ddo_default() `

Verify there is no orphaned 'j' character appearing below the footer.


_docker/drupal/drupal-filesystem/web/themes/custom/rhdp2/templates/system/html.html.twig
```diff
@@ -277,6 +277,6 @@
         _satellite.pageBottom();
       }
     </script>
-    {% endif %}j
+    {% endif %}
```
jordanpagewhite: 😄

PR #3548, opened 2020-04-14
[No Issue] Update to rhel-76.4 version of base image

robpblake:
This commit updates to the rhel-76.4 version of the developer base image which has been updated to incorporate changes in IT PaaS entitlement management when building images in Managed Platform

### Verification Process

* The build should go green
* Log in to the Drupal container and verify the version of Ansible is 2.9 by using `ansible --version` from the command line.


jordanpagewhite:
Thank you again Rob! The build steps are all green, and I verified that the Ansible version is 2.9.6 ✅

PR #3550, opened 2020-04-14
[RHDX-429] Add Event Sessions, Assemblies and Reorganize Fields

ghost:
### JIRA Issue Link
* https://projects.engineering.redhat.com/browse/RHDX-434
* https://projects.engineering.redhat.com/browse/RHDX-436
* https://projects.engineering.redhat.com/browse/RHDX-437
* https://projects.engineering.redhat.com/browse/RHDX-438

### Questions
1. Does a Session need to support multiple languages?
    * **A:** No
2. Will all events have a start and end date and time? If not, we need to define a way to not show the time field on the front end. 
    * **A:** All events should have a required start time. Need to decide what to do with existing events w/o a start time. Perhaps go back and add them manually.
3. Where should "Event Categories" be in the sort order of Purpose Attributes? It's currently at the bottom.
    * **A:** Move to top on Event nodes.

### Verification Process
**Sample Event:** https://developer-preview-3550.ext.us-west.dc.preprod.paas.redhat.com/node/217495/edit

#### RHDX-438 - Add assemblies to Events content type
1. Edit an event node.
2. See that the "Compose" tab now holds only the title and assembly sections.
3. See that the event fields have been moved to the "Fields" tab and reorganized to be easier to enter content.
4. Navigate back to "Compose" tab and add any available assembly to the node and save.
5. View the assemblies on the event node front-end.

#### RHDX-437 - Hide unused fields
1. Edit an event node.
2. Navigate to the "Fields" tab.
3. Note that the following fields are not present:
    * Target product 
    * Data Tags
    * Solution Type
    * Event Format
    * Tile thumbnail

#### RHDX-434 & RHDX-436 - Create sessions and add them to events
1. Edit an event node.
2. Navigate to the "Fields" tab.
3. Scroll down and note that there is a new "Event Category" Vocabulary available in the Purpose Attributes section. Terms will display here once added to the Vocabulary in the Taxonomy section of Drupal.
4. Scroll up to the "Sessions" field.
5. Click the "Add new assembly" button to add a Session assembly to the event.
6. When adding a session, you should be able to enter:
    * Start date and Time
    * End date and time
    * One or more speakers/presenters
    * One language
    * ICS calendar fields


rogeralloymagnetic:
rebuild this please

jordanpagewhite:
rebuild

jordanpagewhite:
rebuild this please

jordanpagewhite:
rebuild this please

jordanpagewhite:
rebuild this please

jordanpagewhite:
rebuild this please

robpblake:
rebuild this please

sjcox-rh:
@rogeralloymagnetic just reviewed the PR!  It's looking good and so excited at all of these changes!

I have some feedback but first wanted to address your questions:

1.  A single session will NOT have to support multiple languages.
2. Not entirely sure but I was wondering if the end time would be shown at all within the frontend. For the most part showing just the start time suffices.
3. Will the event categories purpose attribute section only be shown on event nodes? If so, let's have that category appear at the top of the purpose attributes section. If it has to be shown on every node PA section, then let's have it under BU. I'm assuming for all events that at least one PA under an event categories is required. If that's not the case can we make that a requirement?

**Feedback:**

-   Within the session section, when adding a session, we need to include a way for us to enter in the timezone. Just found out today that each row should showcase the specific timezone, even if it's all the same one. Not sure best way to do that, if it should be included within the time field, or introduce a third field to enter in a timezone.

-  This is minor but that helper text is right on top of the CTA to add another item within the session speaker fields.
Screen Shot 2020-04-16 at 1 55 23 PM

- Another minor thing, but within the session assembly, could we group this whole section under an "Add to Calendar Information" like how it's done within the parent event node? I like how the whole section is collapsed by default.
Screen Shot 2020-04-16 at 1 55 31 PM

Thanks so much!

alistairmc:
rebuild this please

rogeralloymagnetic:
@sjcox-rh Late reply on your review notes. I've updated the main description with your answers to my questions.

1. We talked about this on a call, adding here for reference. All times will be entered in UTC and I've added labels to the fields on the front and back-end to make this clear.
2. It looks like that's a display issue in Beryllium that's outside the scope of this ticket.
3. I had tried to put those in a field group, but it looks like Assembly doesn't support it. Field Groups is actually an external module we use, so since it's not core it takes specific integration.

staceymosier (changes requested):
I walked through the steps and updating events/sessions is working as described. Naming of fields and structure looks great.

I left comments on RHDX-437 and RHDX-438.  Please see those in JIRA and let me know what you think.

rogeralloymagnetic:
@staceymosier
- Issue with RHDX-437 resolved in c29dfc9
- Issue with RHDX-438 resolved in f792bc1

jordanpagewhite:
rebuild this please

staceymosier (approved):
I see the extra fields moved to "Deprecated", and I see the assemblies are rendering full width now without left/right white space.  Thanks!

jordanpagewhite:
This looks ready to merge after my comment is resolved on the katacoda reference from field_sections on the assembly_page node type

jordanpagewhite:
Looks good to me. Thanks for that last change Roger.

PR #3562, opened 2020-04-29
[RHDX-450] Do not load trustarc on local environment

jordanpagewhite:
This makes the following changes to asset/libraries loading: we will now
load trustarc in the rhdp theme (it previously/curently is not), the
JIRA issue collector will load in the at the bottom of the <body>,
trustarc will not load on local environments.

### JIRA Issue Link
* https://projects.engineering.redhat.com/browse/RHDX-450

### Verification Process
* TrustArc scripts/assets do no load on local environments
* TrustArc scripts/assets and functionality continues to behave exactly as it does now on all non-local environments
* TrustArc now loads on the `rhdp` theme (old, paragraphs-based Product pages e.g. RHEL)
* The JIRA issue collector loads at the bottom of the <body>


jordanpagewhite:
rebuild this please

PR #3565, opened 2020-05-04
[RHDX-465] Compact Dynamic Article List Length

jordanpagewhite:
This adds a field to the Compact Dynamic Article List assembly type, the
same field that is on the Dynamic Content Feed, that allows an editor to
configure the number of articles displayed in the list. The current
number of articles displayed is 8. This provides a default of 8, but
allows an editor to set the valuable anywhere between 4 to 10.

### JIRA Issue Link
* https://projects.engineering.redhat.com/browse/RHDX-465

#### Doesn't break existing Dynamic Content Article Lists

Go to the [homepage](https://developer-preview-3565.ext.us-west.dc.preprod.paas.redhat.com/) and verify that the list in the left sidebar, at the very top of the page still renders as expected.

##### Editors can set number of articles per assembly

* Go to the [homepage edit form](https://developer-preview-3565.ext.us-west.dc.preprod.paas.redhat.com/home/edit/) and observe the Number of Posts field
* Set some number of posts between 4 and 10
* Save the field and [view the homepage](https://developer-preview-3565.ext.us-west.dc.preprod.paas.redhat.com/)
* You should see that the number of articles in the list is the number you just set

Ex. 6 posts

Screen Shot 2020-05-04 at 9 50 19 AM

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/modules/custom/rhd_assemblies/src/Plugin/AssemblyBuild/CompactDynamicArticleListBuild.php
+++ b/_docker/drupal/drupal-filesystem/web/modules/custom/rhd_assemblies/src/Plugin/AssemblyBuild/CompactDynamicArticleListBuild.php
@@ -4,14 +4,11 @@ namespace Drupal\rhd_assemblies\Plugin\AssemblyBuild;
 
 use Drupal\Core\Entity\Display\EntityViewDisplayInterface;
 use Drupal\Core\Entity\EntityInterface;
-use Drupal\assembly\Plugin\AssemblyBuildBase;
-use Drupal\assembly\Plugin\AssemblyBuildInterface;
-use Drupal\node\Entity\Node;
-use Drupal\Core\HttpClient;
 
 /**
- * Displays a list of recent content from Wordpress and Drupal with Disqus comments and related topics
- *  @AssemblyBuild(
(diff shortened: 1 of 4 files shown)


_docker/drupal/drupal-filesystem/web/modules/custom/rhd_assemblies/src/Plugin/AssemblyBuild/CompactDynamicArticleListBuild.php
```diff
@@ -4,27 +4,31 @@
...
  */
 class CompactDynamicArticleListBuild extends DynamicContentFeedBuild {
 
+  /**
+   * Build the CompactDynamicArticleList render array.
+   */
   public function build(array &$build, EntityInterface $entity, EntityViewDisplayInterface $display, $view_mode) {
-    $count = 8;
+    $count = $entity->get('field_number_of_posts')->getValue();
+    $count = reset($count)['value'] ?? 8;
```
staceymosier, line 24: Minor feedback:   It would be helpful to add field helper text on the values allowed (minimum: 4, maximum: 10),
and also if we have a default value of 8, then the field should not become required.

I was able to update a rich text assembly on a page and save, without having to update the compact dynamic article list to provide the missing field, however.

PR #3569, opened 2020-05-06
[RHDX-487] Restore assembly audience visibility

staceymosier:
### JIRA Issue Link
https://projects.engineering.redhat.com/browse/RHDX-487

### Verification Process
This restores the functionality that was found in rhdp theme with minor changes.

- Create an assembly that has an Audience Selection field, and set the audience selection to Authenticated.
- Create a second assembly that has the audience selection to Unauthenticated.
- View the assembly at /assembly/##/view, or attach the assembly to a supported content type and view the node.  
When unauthenticated (via Keycloak), verify that the content for the unauthenticated audience shows. Also, that the content for the authenticated audience doesn't.
- When authenticated, check that the opposite occurs. 

You can also view this here
https://developer-preview-3569.ext.us-west.dc.preprod.paas.redhat.com/test/audience-selection/


jordanpagewhite:
rebuild this please

jordanpagewhite:
This looks ready to go once we jQuery .detach() or vanilla JS .removeChild() the element that should not be visible to the user.

jordanpagewhite:
rebuild this please

PR #3581, opened 2020-05-27
[NO ISSUE] Resolve download_url fromUri error

jordanpagewhite:
This resolves an error that arose on our new local environments where
cheat sheet cards are being rendered, and it results in an error because
the $download_url is trying to be built using Url::fromUri() without
passing a protocol. This resolved this error for me locally.

### JIRA Issue Link
* n/a

### Verification Process

* You can view a page with a cheat sheet card, and you do not see an error like "InvalidArgumentException: The URI '/download-manager/.....' in invalid. You must use a valid URI scheme in Drupal\Core\Url:fromUri()....


jordanpagewhite:
rebuild this please

rogeralloymagnetic:
rebuild this please

PR #3584, opened 2020-06-04
[RHDX-560] OnPageNavigation should load by revision id

staceymosier:
### JIRA Issue Link
[RHDX-560](https://projects.engineering.redhat.com/browse/RHDX-560)

This PR adds the `['#cache'][tags]` to the content.nav render array for the Jump to Nav assembly. It also adds Patternfly inline list classes to the ul.

I also cleared phpcs linting warnings for files in rhd_assemblies/src/Plugin/AssemblyBuild/.
I did not clear phpcs linting warnings for the other files in /src/ at this time.

### Verification Process
Check topic and assembly pages that have a Jump To Nav assembly.
Verify the list renders correctly and the links jump down to the referenced assembly. (The links are anchor links to content further down on the page.) 

Assemblies are linked to Pages by entity reference revision.  This means that 2 different pages can reference an assembly, at different revision of that assembly.  The workflow to test/support this varies, but I think it can be verified by:

1. Look for a page that has a rich text assembly that is already published.
2. Create a new page, link that same rich text assembly to it and check 'create new revision' on the assembly.  Make changes to that assembly and verify it does not show up on the first page.


staceymosier:
rebuild this please

staceymosier:
Rebuild this please

staceymosier:
rebuild this please

staceymosier:
Rebuild this please.

staceymosier:
rebuild this please.

staceymosier:
Rebuild this please.

_docker/drupal/drupal-filesystem/web/modules/custom/rhd_assemblies/src/Plugin/AssemblyBuild/OnPageNavigationBuild.php
```diff
@@ -63,12 +74,19 @@ public function build(array &$build, EntityInterface $entity, EntityViewDisplayI
...
+              'pf-c-list', 'pf-m-inline ',
+              'rhd-u-no-bullet',
+            ],
+          ],
         ];
       }
     }
 
-    $build['#cache']['tags'][] = $build['#parent']['entity']->getEntityTypeId() . ":" . $build['#parent']['entity']->id();
+    $cacheableMetadata->applyTo($build);
```
jordanpagewhite, line 88: Just for reference / my own sanity in the review, this is what we're adding to the `#cache`. This is what we would expect. I am not 100% certain that we _need_ to add assemblies that are placed above the OnPageNavigation assembly, in this case the https://localhost/assembly/32715/view, but I don't think that it really hurts necessarily. Do you have thoughts @staceymosier about whether or not it is worth looking into excluding assemblies that are not rendered in the OnPageNavigation in this PR?

```
array(5) {
  ["tags"]=>
  array(7) {
    [0]=>
    string(14) "assembly:32715"
    [1]=>
    string(14) "assembly:33055"
    [2]=>
    string(14) "assembly:33075"
    [3]=>
    string(14) "assembly:33095"
    [4]=>
    string(14) "assembly:33155"
    [5]=>
    string(14) "assembly:33175"
    [6]=>
    string(14) "assembly:33555"
  }
  ["contexts"]=>
  array(0) {
  }
  ["max-age"]=>
  int(-1)
  ["keys"]=>
  array(4) {
    [0]=>
    string(11) "entity_view"
    [1]=>
    string(8) "assembly"
    [2]=>
    string(5) "33035"
    [3]=>
    string(7) "default"
  }
  ["bin"]=>
  string(6) "render"
}
```

jordanpagewhite:
Everything looks great. I will just wait to hear back on my one comment, but we won't necessarily need to / want to make further changes.

alistairmc:
rebuild this please

_docker/drupal/drupal-filesystem/web/modules/custom/rhd_assemblies/src/Plugin/AssemblyBuild/OnPageNavigationBuild.php
staceymosier: I pushed a change to include the parent entity (node) as a cache tag. 
I think keeping all the available assemblies as cache tags is necessary, in the event that they are rearranged on the node.  Although right now the node itself can suffice as the only cache tag,  the rest of the assemblies here are necessary in the case that the 'assembly update triggers node update' design changes in the future.

staceymosier:
Rebuild this please.

alistairmc:
rebuild this please

PR #3585, opened 2020-06-09
[RHDX-409] Add upcoming_event_cards View to events_collection assembly.

ghost:
### JIRA Issue Link
* https://projects.engineering.redhat.com/browse/RHDX-409

**NOTES:**
- The View used to create the card Display Option for this assembly uses the Event "start_date" field to sort and filter by.
- It will default to using the Event "start_date" in the card display unless one or more sessions are entered on the event. Once sessions are added to an Event, only session dates are displayed.
- It will show events and sessions from "today" forward.
- The event will fall off after today's date. If there is a session date later than the Event date, the card will fall off based on the Event date. This means that for now, Event dates will need to be updated to the next upcoming session for that event.

### Verification Process
1. Edit DevNation node: https://developer-preview-3585.ext.us-west.dc.preprod.paas.redhat.com/devnation/edit
2. Create a new "Events Collection" assembly.
3. Choose the "Upcoming events - (cards)" Display Option
3. Add some Event Categories to filter by.
4. Keep or change the "Limit Results" value.
5. Save the assembly and node.
6. View the assembly on https://developer-preview-3585.ext.us-west.dc.preprod.paas.redhat.com/devnation
7. **Bonus Points:** Play with configurable result limits and display options.
8. **More Bonus Points:** Add multiple sessions to an event to see it use the session dates instead of the event date in the bottom author section of the card.

### Screenshots
Screen Shot 2020-06-10 at 1 03 11 PM
Screen Shot 2020-06-10 at 1 10 17 PM

### DIFF
--- a/_docker/drupal/drupal-filesystem/web/config/sync/core.entity_form_display.assembly.events_collection.default.yml
+++ b/_docker/drupal/drupal-filesystem/web/config/sync/core.entity_form_display.assembly.events_collection.default.yml
@@ -5,6 +5,7 @@ dependencies:
   config:
     - assembly.assembly_type.events_collection
     - field.field.assembly.events_collection.field_audience_selection
+    - field.field.assembly.events_collection.field_card_limit
     - field.field.assembly.events_collection.field_display_option
     - field.field.assembly.events_collection.field_drupal_term_filter
     - field.field.assembly.events_collection.field_show_filter
@@ -24,6 +25,12 @@ mode: default
 content:
   field_audience_selection:
     weight: 2
+    settings: {  }
+    third_party_settings: {  }
+    type: options_select
+    region: content
+  field_card_limit:
+    weight: 27
     settings: {  }
     third_party_settings: {  }
     type: options_select
--- a/_docker/drupal/drupal-filesystem/web/config/sync/core.entity_view_display.assembly.events_collection.default.yml
+++ b/_docker/drupal/drupal-filesystem/web/config/sync/core.entity_view_display.assembly.events_collection.default.yml
@@ -5,6 +5,7 @@ dependencies:
   config:
     - assembly.assembly_type.events_collection
     - field.field.assembly.events_collection.field_audience_selection
+    - field.field.assembly.events_collection.field_card_limit
     - field.field.assembly.events_collection.field_display_option
     - field.field.assembly.events_collection.field_drupal_term_filter
     - field.field.assembly.events_collection.field_show_filter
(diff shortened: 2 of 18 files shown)


rogeralloymagnetic:
rebuild this please

maryshak1996:
@rogeralloymagnetic the only issues that I see are:
1) the card heights should all adjust to be the height of the tallest card
2) for the cards that don't have a date yet, could we have a placeholder date tag that says 'Dates: To be determined' ?
Screen Shot 2020-06-16 at 1 20 12 PM

staceymosier (changes requested):
There are 3 phpcs warnings for rhdp2.theme
 412 | [x] Expected 1 space after "="; 0 found
 435 | [x] Use "elseif" in place of "else if"
 449 | [x] Whitespace found at end of line

The default value for the field_card_limit in config is 4, but the fallback in the php is 0? 
Can we default to 4 instead of unlimited (if I understand that correctly).

Minor: The event cards do not keep a consistent height; Not sure if that is part of the design manual or variance in content. See [221720](https://developer-preview-3585.ext.us-west.dc.preprod.paas.redhat.com/node/221720/).

rogeralloymagnetic:
@maryshak1996 
1. The card heights are resolved.
2. What you were seeing was a programming error with the date display. Events won't be displayed if there is no date. Now, the logic that causes the date field and the event itself to fall off of the view are aligned so you won't see an event with no date.

@staceymosier 
1. PHPCS errors are resolved.
2. As discussed on Slack, the total posts need to accept zero as a valid option so that default is purposeful.
3. Duplicate resolved in Mary's notes.

Screen Shot 2020-06-17 at 12 58 23 PM

staceymosier (approved):
I edited an event to remove it's reference to the "registration page", and the card builds with the Event alias instead.   Everything looks great and is updating after changes are made to the event data.

rogeralloymagnetic:
Approval from Mary was given on Slack. Merging this one.

maryshak1996:
hey @rogeralloymagnetic ... I realized that I totally missed this when I reviewed, but the event title is getting duplicated in the link title and the description below it in the cards, which is a little bit redundant. I just filed this bug, to have that body section swap to listing the speakers instead of a repeat of the event title: https://projects.engineering.redhat.com/browse/RHDX-625
Again, sorry about not noticing this before!

rogeralloymagnetic:
@maryshak1996 I think this is a content issue with what's entered in the "Short Description" field. Can you please double check?

maryshak1996:
@rogeralloymagnetic I just checked, and you're right -- the content creator of the actual event node is listing the name in the short description. But, if possible, I still think that it would be better that that area reference the 'Presenter' instead of the 'short description' to display right there, since the current hard-coded cards list the speaker names, not a description.

